PR #2, opened 2023-11-07
feat: allow users to define custom session factory + types

danielroe:
## Questions

1. should `onSuccess` be fully optional (as in this implementation)?
1. would we instead like to require `onSuccess` (e.g. for redirect) _but also_ always use the defined session factory for setting user session?
1. we should investigate type issues with `@nuxt/ui-pro` in playground (run `pnpm test:types:playground` to see them)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,8 +65,8 @@ jobs:
       - name: Run test suite
         run: pnpm test
 
-      # - name: Test types
-      #   run: pnpm test:types
+      - name: Test types
+        run: pnpm test:types
 
       # - name: Test playground types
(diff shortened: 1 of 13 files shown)


Atinux:
> should onSuccess be fully optional (as in this implementation)?

We should make it required as best practice and avoid doing hidden stuff

> would we instead like to require onSuccess (e.g. for redirect) but also always use the defined session factory for setting user session?

Not sure to understand the defined session factory you mean

For UI Pro types, I think it comes from Nuxt 3.8.1, @benjamincanac is working on it

danielroe:
so I guess the question is whether we just run the session handler in `defineSession` + automatically do `setUserSession`?

PR #4, opened 2023-11-08
chore: Set `useSession` `SessionDataT` type to `UserSession`

Gerbuuun:
Instead of asserting the session data, pass the `UserSession` type as a generic into the h3 `useSession`.
Also, use the `SessionConfig` instead of `any`.
(Although `ts-ignore` is used so it doesn't really do anything, it is more correct)


Atinux (approved):
Looks good to me but want @danielroe expert TS eye :D

danielroe (approved):

PR #7, opened 2023-11-09
feat: added discord auth provider

justserdar:
Created a Discord provider based on the spotify provider.

The Discord provider is useful when you want to connect your Nuxt 3 project with your Discord Community.
Have information about your users displayed in your codebase or just simply create cool widgets or things visitors or users can see.

Screenshot_2023_11_09-1

The auth flow requires a HTTPS redirect uri, I have it working using NGROk locally.
Discord provides a developer section where you can setup your applications and get your token information.
The user will have to setup their oauth settings at Discord like any other platform.

I found a challenge when introducing the redirect_uri as a configurable variable in the config because it needs to redirect back to 'https://your-domain.com + /auth/discord'
Edit: Improved it with the domain variable


Atinux:
Thank you for the PR @justserdar 

You don't need to use ngrok actually, you can simply have HTTPS with Nuxt in development with `npx nuxi dev --https`

We should not have a variable for the `domain` IMO like the others Oauth providers we have.

justserdar:
Hey!

I removed the domain usage and went for the existing request path route as per the other oauth providers.
Thank you about the tip with --https!

PR #8, opened 2023-11-09
feat: added Microsoft as oauth provider

jfrelik:
Added Microsoft as an oauth provider

image


curlydoodle:
Some Microsoft logins for government instances use https://login.microsoftonline.us

https://graph.microsoft.us/v1.0/me 
https://portal.azure.us 

maybe having a domain .env variable would make this work with government and non government Microsoft instances.

jfrelik:
Never heard of it maybe because it is only available for US GOV
https://learn.microsoft.com/en-us/azure/azure-government/documentation-government-aad-auth-qs

I would prefer to do this via optional env variable with a boolean like
NUXT_OAUTH_MICROSOFT_GOVERNMENT=true

curlydoodle:
Yea US Gov instance uses all .us domains.

same goes for their graph api

yep, having that optional .env would work great

Atinux:
Agree by adding `NUXT_OAUTH_MICROSOFT_US_GOV` env variable

jfrelik:
@Atinux @curlydoodle,

Added `NUXT_OAUTH_MICROSOFT_US_GOV` as an env variable
When env is not set auth works correctly by using .com domain, if declared and set to true it resolves to .us
Cannot test this without Azure Gov instance

Atinux:
Thanks!
Happy to resolve the conflicts @jfrelik ?

jfrelik:
@Atinux 

Conflicts resolved should be good to go :)

curlydoodle:
From my testing, I believe Microsoft forwards to the correct domain based on the tenant id that is supplied. But I still think how @jfrelik implemented this is good to ensure we explicitly define which instance to use. I didn't get to test all the way through since its asking me to grant access permission but that could be due to me reusing an existing client id.

`needs permission to access resources in your organization that only an admin can grant`

RihanArfan:
Just want to point out there's more instances than just global and US gov. From what I can briefly see, there's global, US Gov, China, Germany and potentially more. https://learn.microsoft.com/en-us/entra/identity-platform/authentication-national-cloud

I think we should default to global, and have an environment variable to override instance ([which is what Microsoft's auth SDKs call this](https://learn.microsoft.com/en-us/entra/identity-platform/web-app-tutorial-02-prepare-application?tabs=visual-studio#define-the-platform-and-urls:~:text=Instance%20%2D%20The%20authentication%20endpoint.%20Check%20with%20the%20different%20available%20endpoints%20in%20National%20clouds.)).

Atinux:
Then what about letting the user configure the authorization url and tokens url like for GitHub?
https://github.com/Atinux/nuxt-auth-utils/blob/1b296e2bf841e6288ef7ce6d2a19892f20912ad1/src/runtime/server/lib/oauth/github.ts#L32-L42

jfrelik:
That can be the solution, although looking at some other projects like Appwrite they only support global .com domain

https://github.com/appwrite/appwrite/blob/7766279025424f390032c8bee558913725e1a402/src/Appwrite/Auth/OAuth2/Microsoft.php#L65

RihanArfan:
I suggest we follow what Microsoft's SDKs does - https://learn.microsoft.com/en-us/entra/identity-platform/tutorial-v2-nodejs-webapp-msal#add-app-registration-details

NUXT_OAUTH_MICROSOFT_CLOUD_INSTANCE="Enter_the_Cloud_Instance_Id_Here" # cloud instance string should end with a trailing slash - default to https://login.microsoftonline.com/
NUXT_OAUTH_MICROSOFT_TENANT_ID # "common" | "consumers" | string
NUXT_OAUTH_MICROSOFT_CLIENT_ID
NUXT_OAUTH_MICROSOFT_CLIENT_SECRET
NUXT_OAUTH_MICROSOFT_SCOPES # scopes if granting access to access other APIs or endpoints on Graph
NUXT_OAUTH_MICROSOFT_GRAPH_API_ENDPOINT # default to https://graph.microsoft.com/ - some people might want to use beta endpoint

The token and and authorization URL path remain the same, the with the only part changing is the instance and tenant IDs, so I don't think it should be configurable.
- https://<instance>/<tenant>/oauth2/v2.0/authorize
- https://<instance>/<tenant>/oauth2/v2.0/token

Atinux:
I don't think we need an env variable each time since it can configure it with the `config` prop of `oauth.microsoftEventHandler` or the runtimeConfig.

```
NUXT_OAUTH_MICROSOFT_CLIENT_ID=
NUXT_OAUTH_MICROSOFT_CLIENT_SECRET=
NUXT_OAUTH_MICROSOFT_TENANT_ID=
```

Then we could give for configuration:

```ts
export default oauth.microsoftEventHandler({
  config: {
    authorizationURL: '', // custom url
    tokenURL: '', // custom url
    userURL: '', // custom url
    scope: []
  },
  // ...
})
```

PR #12, opened 2023-11-10
feat: added simple pkce and state checks for auth0

Azurency:
I don't if that's something that something in the scope of a "Minimalist Authentication module" but I added simple PKCE and state checks in the auth0 provider for extra security.

The crypto methods and check logic could easily be extracted and reused in other oauth provider. I think that Google and Spotify accept the PKCE flow and Google and Twitch both accept a state check.


justserdar:
The other providers could definitly benefit from a universal util that allows state checks and a pkce check.
If we can get this into like a importable util, i'd happily add it to the Discord and LinkedIn provider too.

sifferhans:
You should probably use `uncrypto` instead of `crypto` here, as `crypto` might not be available in all server runtimes 😁

Atinux:
I would indeed leverage https://github.com/unjs/uncrypto 

Also here we use `randomUUID` to create a random string: https://github.com/nuxt/nuxters/blob/d77b0e92092c39710aeb36d604d85b886bdc87b7/server/routes/connect/%5Bprovider%5D.get.ts#L12

Azurency:
Tanks for the feedback 👍, I updated the checks to use `uncrypto`. 

I also moved the logic in a separate util file so that other providers can simply call
```
const authParam = await checks.create(event, config.checks)
```
to get extra query param to be passed in the authorization request and
```
const checkResult = await checks.use(event, config.checks)
```
to verify the checks and get back the code_verifier if applicable.

src/runtime/server/utils/security.ts
```diff
@@ -0,0 +1,117 @@
...
+   * @param event, H3Event
+   * @param checks, OAuthChecks[] a list of checks to create
+   * @returns Record<string, string> a map of check parameters to add to the authorization URL
+   */
+  async create(event: H3Event, checks?: OAuthChecks[]) {
+    const res: Record<string, string> = {}
+    if (checks?.includes('pkce')) {
+      const pkceVerifier = generateCodeVerifier()
+      const pkceChallenge = await pkceCodeChallenge(pkceVerifier)
+      console.log('pkceVerifier', pkceVerifier)
```
maximilianmikus, line 69: console.log leftover

src/runtime/server/utils/security.ts
```diff
@@ -0,0 +1,117 @@
...
+      console.log('pkceChallenge', pkceChallenge)
+      res['code_challenge'] = pkceChallenge
+      res['code_challenge_method'] = 'S256'
+      setCookie(event, 'nuxt-auth-util-verifier', pkceVerifier, { maxAge: 60 * 15, secure: true, httpOnly: true, sameSite: 'lax' })
```
maximilianmikus, line 73: I think the cookie settings should configurable or reuse the cookie settings from the module

maximilianmikus:
I would like to build upon this PR and add a generic OIDC provider. I already started but these utils would be super helpful for supporting pkce flow.

src/runtime/server/utils/security.ts
Azurency: I didn't reuse the cookie settings because they were under the session key, I don't know if it would be confusing to reuse that or not. But I agree that a shared cookie config somewhere would be nice.

maximilianmikus: I agree it would be a bit confusing. Maybe an optional config for pkce cookie?

src/runtime/server/utils/security.ts
```diff
@@ -0,0 +1,115 @@
...
+      res['code_verifier'] = pkceVerifier
+    }
+    if (checks?.includes('state')) {
+      const stateInCookie = getCookie(event, 'nuxt-auth-util-state')
+      setCookie(event, 'nuxt-auth-util-state', '', { maxAge: -1 })
+      if (checks?.includes('state')) {
+        if (!state || !stateInCookie) {
+          const error = createError({
+            statusCode: 401,
+            message: 'Auth0 login failed: state is missing'
```
maximilianmikus, line 100: This message is still specific to Auth0

src/runtime/server/utils/security.ts
```diff
@@ -0,0 +1,115 @@
...
+          })
+          throw error
+        }
+        if (state !== stateInCookie) {
+          const error = createError({
+            statusCode: 401,
+            message: 'Auth0 login failed: state does not match'
```
maximilianmikus, line 107: This message is still specific to Auth0

src/runtime/server/utils/security.ts
Azurency: I added a cookie setting (runtimeConfig), under `nuxtAuthConfig.security.cookie`

itpropro:
Hey @Azurency,
the current implementation in the security utils file is not RFC 7636 compliant. There are some issues with the generation of the verifier and the challenge, the verifier is saved server side, which should be the challenge and the hash validation doesn't take place.
I could create a PR to adjust your implementation to the spec, but I have some questions:

- What is the target of the current handler implementation for PKCE? There is currently no identity provider in this repo that would need a verification of the PKCE challenge or verifier on the server side, as we are not issuing tokens to the user
- What is the target of the state check? The referenced auth0 documentation refers to SPA flow, but as we are using Nuxt SSR, we have a confidential client scenario. For simple CSRF protection, setting a fixed CSRF header value would be enough, as this would force all client to make preflight requests

itpropro:
> @itpropro ` There is currently no identity provider in this repo that would need a verification of the PKCE challenge`
> 
> As @Azurency states in the PR description: `The crypto methods and check logic could easily be extracted and reused in other oauth provider. I think that Google and Spotify accept the PKCE flow and Google and Twitch both accept a state check.` Also #25 relies on this
> 
> @itpropro I think it would be great if you could improve the implementation.

Hey,
I think I wasn't clear with my statement. I didn't mean that there is no authentication provider that would use PKCE, but they would validate the verifier on their end, so that there is no need to verify at the Nuxt server side. I was just talking about the right place and level of PKCE in the implementation as it's about having a "PKCE client" no a PKCE server implementation.
I think it makes sense to have some kind of pipeline where the requests go through and the supported/configured features are applied (like PKCE, Nonce, State, etc.).

amandesai01:
Thanks @itpropro I was exactly getting confused by the same issue all along while reading code changes and discussion.

What I would suggest is adding those utils on client side, there by allowing this module to work with SPAs and `nuxt generate`. Then we can have a separate set of clients which work client side. I think it should be a separate discussion. Work done in this PR is extemely valuable but should be ported to client side.

PR #14, opened 2023-11-11
feat: refactor login buttons to use dropdown

arashsheyda:
With more providers getting added, this design allows better organization and presentation.

https://github.com/Atinux/nuxt-auth-utils/assets/38922203/12324c55-2f8f-45ed-9507-3a9a61703a73


Atinux:
I was actually thinking of doing this or directly use card in the main page for each provider

You read my mind :D

arashsheyda:
I'm glad we're thinking the same :D
Card will look good too actually, maybe we can add both? or dropdown is enough?

Atinux:
Let's keep the dropdown for now 😀 

I will have to deploy the playground also to have a nice demo.

But need to create all the OAuth apps 😆

arashsheyda:
Yeah that'd be nice, also I think some docs would be helpful, I can work on that if you think it's a good idea

Atinux:
More docs than what we have on the readme? What are you thinking of?

arashsheyda:
Oh sorry my last comment is not that clear now that I read it :d I think we can add the providers developer links(similar to what you wrote in `nuxt-todo-edge` setup for github)

and with a website maybe we can add examples page so user can try the features out

PR #18, opened 2023-11-13
fix(oauth): add generic OAuthConfig type

sifferhans:
I noticed a separate `OAuthConfig` interface was defined for every provider, where only the config type was differing between them.

I thought maybe it might make sense to have a generic `OAuthConfig` type shared between them.
If it doesn't make sense, is there a reason why?


sifferhans:
Also, what is the best way to import the `OAuthConfig` type in the providers? I used a relative path for now, but I'm aware that might cause problems

Atinux:
Happy to quickly look at it @danielroe about where the best place to have a generic type?

danielroe (approved):

Atinux (approved):

PR #27, opened 2023-12-01
fix(auth0): add audience to auth0 runtime config types

LeoMo-27:
This pull request aims to add `audience` to the auth0 runtime config types, since it should be possible to add it. Right now it says it's not assignable if it's the first time you are writing the config, which is a little bit misleading.

image


Azurency:
Oh, you're right, it was removed here https://github.com/Atinux/nuxt-auth-utils/pull/6/commits/2fbc66e606c8994c0b403c58c1a03944dc4fec8c

PR #36, opened 2024-01-07
feat: added aws cognito provider

dvh91:
added aws cognito provider


Atinux:
Thank you for the PR @dvh91 !

Do you think we can have the `emailRequired: boolean` option to make sure we get back an email or it's always the case?

PR #40, opened 2024-01-21
fix: replace encoded space characters with regular spaces

ahmedrangel:
I've been using Twitch auth and tried to incorporate certain scopes. However, I got a scope error during authentication and noticed that the encoded spaces were being re-encoded in the query. I also found several providers had this same problem and made these changes.


Atinux:
Thank you!

PR #46, opened 2024-02-05
Configurable server handler endpoints

maximilianmikus:
This PR adds a config option to make the server handler endpoints configurable. 
You can now configure if you want to use the built-in server handlers for fetching and deleting a session (default). Or if you want to define your own endpoints and implement your custom logic there.

In my case it was necessary because the hooking system was not working how I needed it to be (every mutation to the session object saves new session data). I think adding an option to make them configurable does not hurt much and makes the module much more flexible.

Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,16 +64,20 @@ Define custom server handler endpoints. Makes it possible to use your own custom
 
 ```js
 export default defineNuxtConfig({
-  auth: {
-    serverHandler: {
-      getSession: {
-        route: '/api/custom/auth/session',
-        method: 'get'
-      },
-      deleteSession: {
-        route: '/api/custom/auth/session',
-        method: 'delete'
-      },
+  runtimeConfig: {
+    public: {
+      auth: {
+        serverHandler: {
+          getSession: {
+            route: '/api/custom/auth/session',
+            method: 'get'
+          },
+          deleteSession: {
+            route: '/api/custom/auth/session',
+            method: 'delete'
+          },
+        }
+      }
     }
   }
 })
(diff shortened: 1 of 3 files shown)


Atinux:
Thanks for the PR @maximilianmikus 

I thought about it too but IMO the hooks should do the work (each new config option is more complex logic to handle).

Can you better explain to me at first why the hooks was not working as intented? Do you have an example?

maximilianmikus:
I am implementing saving session data in redis. I tried to use the hook system (fetch) but every mutation to the session object saves new session data. So in the hook I retrieved the data from redis. I added the data to the session object so it gets returned in the response. I run into the issue of the cookie being to big because the data is saved to the cookie. I want to use the cookie just for the cookie.id and have all data in redis (we have a lot of data and its too big for the cookie).

Atinux:
Hey @maximilianmikus 

The `fetch` hook does not update the session data, thus not updating the cookie directly: https://github.com/Atinux/nuxt-auth-utils/blob/main/src/runtime/server/api/session.get.ts#L7

I made a test and can confirm that this code: https://github.com/Atinux/nuxt-auth-utils/blob/main/playground/server/plugins/session.ts#L5-L7

Does not update the cookie but extend the session data returned for the Vue part (`useUserSession()`)

maximilianmikus:
thank you @Atinux for checking! I don't know if we did something wrong, we just noticed that after adding data from redis to the session (via the `fetch` hook) the cookie size in the frontend was increased significantly. Will try to reproduce it in a sandbox.

Atinux:
Closing in the meantime :)

Issue #38, opened 2024-01-16
Better typing of UserSession


Gerbuuun:
### My problem
It has been bugging me for a while that the types of this module do not work at all.

Lets say this is your `auth.d.ts`:
```ts
declare module '#auth-utils' {
  interface UserSession {
    user: {
      id: number
      name: string
      email: string
    }
    loggedInAt: number
  }
}
```

On the client-side, all properties of the `useUserSession` composable are of type `any` (only in the playground it works)

![Scherm­afbeelding 2024-01-16 om 14 00 54](https://github.com/Atinux/nuxt-auth-utils/assets/12988935/427595dc-51a3-40ba-abed-31c9c6761d4f)

On the server-side, `UserSession` is defined except for the `user` object.
Its type will always be `{} | undefined` regardless of what is specified in the `auth.d.ts` (see issue #31 )
And when checking for a session using `requireUserSession` the returned `user` object is still possibly `undefined` even though we just checked and required it to not be.

### I have been tinkering and came up with the following:

Instead of looking for a specific property in the session data (`user` in this case), we check if there are any properties at all.
```ts
export async function requireUserSession(event: H3Event) {
  const userSession = await getUserSession(event)

  if (Object.keys(userSession).length === 0) {
    throw createError({
      statusCode: 401,
      message: 'Unauthorized'
    })
  }

  return userSession
}
```
This way you can define anything in the `UserSession` interface and the type works (this uses the playground's `auth.d.ts`)

![Scherm­afbeelding 2024-01-16 om 14 22 30](https://github.com/Atinux/nuxt-auth-utils/assets/12988935/9c65eb16-a797-4f54-9ce4-ab409499b08c)

On the client-side I made the following changes:
- `loggedIn` is now computed based on wether there exist properties on the session object
- There is no longer a `user` getter as this property is no longer defined but now the typing is at least correct.
```ts
const useSessionState = () => useState<UserSession>('nuxt-session', () => ({}))

export const useUserSession = () => {
  const sessionState = useSessionState()
  return {
    loggedIn: computed(() => Object.keys(sessionState.value).length > 0),
    session: sessionState,
    fetch,
    clear
  }
}
```

I would like to hear feedback from you guys

Atinux:
Thanks for your issue @Gerbuuun 

I do like having the `user` directly and separated from the session, I think this is just a type issue that we can fix. Sadly I am not a TS expert and the help if more than welcome!

BayBreezy:
[...]

PR #53, opened 2024-02-07
fix: improved types for `useUserSession()`

harlan-zw:
Issue: https://github.com/Atinux/nuxt-auth-utils/issues/38

By explicitly typing the `useUserSession()` we fix the types not being exported correctly. 

I tried to also fix the type augmentation with it, my PHPStorm was freaking out on types though so was hard to see if its fixed.

As far as I could test, if you augment `UserSessionUserData` the `user` should have the right data. Maybe worth confirming on your end.


Atinux:
Closing in favour of #54

PR #54, opened 2024-02-12
fix: UserSession user type augmentation

### LINKED ISSUE #31: Cannot augment `user` in UserSession
The playground project tries to augment `UserSession`:
https://github.com/Atinux/nuxt-auth-utils/blob/e344c9880491346ab73165ac4439c129b13ec15d/playground/auth.d.ts#L3

But `user` is already declared here:

https://github.com/Atinux/nuxt-auth-utils/blob/e344c9880491346ab73165ac4439c129b13ec15d/src/runtime/types/session.ts#L2

🤔I don't think it is possible to augment an already declared field of an interface.

Type errors in playground:
image

### LINKED ISSUE #38: Better typing of UserSession

Gerbuuun:
Closes #38
Closes #31

It seems like the `user` type inside the `UserSession` interface cannot be augmented directly so I created a separate `User` interface. This does work.

Fixing the possibly `undefined` after using `requireUserSession` is achieved by setting the type explicitly to be defined. (we can do this because we check it to be defined right before returning it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,10 +112,15 @@ You can define the type for your user session by creating a type declaration fil
 
 ```ts
 declare module '#auth-utils' {
+  interface User {
+    // Add your own fields
+  }
+
   interface UserSession {
-    // define the type here
+    // Add your own fields
   }
 }
+
 export {}
 ```
 
--- a/playground/auth.d.ts
+++ b/playground/auth.d.ts
@@ -1,17 +1,18 @@
 declare module '#auth-utils' {
+  interface User {
+    spotify?: any
+    github?: any
+    google?: any
+    twitch?: any
+    auth0?: any
+    microsoft?: any;
+    discord?: any
+    battledotnet?: any
+    keycloak?: any
+    linkedin?: any
+  }
+
   interface UserSession {
-    user: {
-      spotify?: any
-      github?: any
-      google?: any
-      twitch?: any
-      auth0?: any
-      microsoft?: any;
-      discord?: any
-      battledotnet?: any
-      keycloak?: any
-      linkedin?: any
-    }
     extended?: any
     loggedInAt: number
   }
--- a/src/runtime/server/utils/session.ts
+++ b/src/runtime/server/utils/session.ts
@@ -3,7 +3,7 @@ import { useSession, createError } from 'h3'
 import { defu } from 'defu'
 import { createHooks } from 'hookable'
 import { useRuntimeConfig } from '#imports'
-import type { UserSession } from '#auth-utils'
+import type { User, UserSession } from '#auth-utils'
 
 export interface SessionHooks {
   /**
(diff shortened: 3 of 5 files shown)


Atinux:
What do you think of it @harlan-zw

Since you also opened #53

harlan-zw:
Seems like it's doing the same as mine, does it generate the correct implicit types for `useUserSession` with this change? Can we close one the PRs if they're accomplishing the same thing

Also, FWIW breaking change of types, the types are already broken and users would need to force type augmentation to use the module as is anyway.

Gerbuuun:
Seems like the return type of `requireUserSession` is compiled to `any` (and all other things as well).
So explicitly typing like @harlan-zw did is also required.
This is on me for forgetting to test the build.

His changes, including the one described below should work.
```ts
export async function requireUserSession(event: H3Event): Promise<UserSession & { user: User }> {
  const userSession = await getUserSession(event)

  if (!userSession.user) {
    throw createError({
      statusCode: 401,
      message: 'Unauthorized'
    })
  }

  return userSession as UserSession & { user: User }
}
```

I'll push the fixes to this branch.
You can decide to merge his branch or mine as it is his work. (be sure to add the changes described above)

PR #56, opened 2024-02-17
feat: add authorizationParams in oauth config

### LINKED ISSUE #22: Add option for additional query parameters
Would be great if the config would have a key like `authorization_params` where we can add any optional parameter the provider expects.

In my example I would like to add `?access_type=offline` to the `authorizationURL` for Google (So that it returns also the refresh token). Having one general `authorization_params` object would avoid having to add different parameters for each provider manually. 

Ideally I would have something like:

```js
export default oauth.googleEventHandler({
	config:{
		authorization_params: {  // Something like this would work perfectly
		    access_type:"offline"
		},
		scope: [
			'https://www.googleapis.com/auth/userinfo.email',
			'https://www.googleapis.com/auth/tasks.readonly' // This is great!
		]
	},
	async onSuccess(event, { user, tokens }) {
		console.log(tokens.refresh_token) // Save this in DB
		await setUserSession(event, { user, loggedInAt: Date.now() })
	  	return sendRedirect(event, '/')
	}
})

```

From my understanding this is currently impossible, right?

### LINKED ISSUE #43: Utils does not work with Azure SWA
We are trying to use this with Azure Static Web App, but the redirect url is wrong.

From what i can see it gets the redirectUrl from `getRequestUrl(event).href`, this does not work with SWA since the hostname refers to an internal url for the function app.

You can see an example output here:
https://ambitious-dune-01c6e3503-preview.westeurope.4.azurestaticapps.net/api/info

Not sure what the best approach for this would be. X-Forward-For header contains only IP Adresse (i assume this is internal infrastructure at Microsoft). SWA does have a special header called 'x-ms-original-url' that contains the original url, it seems very specific for SWA and would only apply to apps running in Azure SWA. 

One possible solution would be to allow a configured redirectUrl (but its not the best solution IMO, in case we have multiple domains pointing to the same instance), another would be to create a custom handler for apps running in SWA.

Any suggestions on the best way to handle this?

Atinux:
Resolves #22
Resolves #43 by overwriting `redirect_uri` in `authorizationParams`

This pull request introduce `authorizationParams` option in each OAuth provider, example:

```ts
export default oauth.googleEventHandler({
  config: {
    authorizationParams: {
      access_type: 'offline'
    }
  },
  async onSuccess(event, { user }) {
    // ...
  }
})
```

The `authorizationParams` can overwrite the generated params we provide and this can be useful in case such as #43 

```ts
export default oauth.microsoftEventHandler({
  config: {
    authorizationParams: {
      redirect_uri: 'https://my-custom-redirect-uri'
    }
  },
    async onSuccess(event, { user }) {
      // ...
    }
})
```

### DIFF
--- a/playground/app.vue
+++ b/playground/app.vue
@@ -3,67 +3,67 @@ const { loggedIn, user, session, clear } = useUserSession()
 
 const providers = computed(() => [
   {
-    label: session.value.user?.github?.login || 'GitHub',
+    label: session.value.user?.github || 'GitHub',
     to: '/auth/github',
     disabled: Boolean(user.value?.github),
     icon: 'i-simple-icons-github',
   },
   {
-    label: session.value.user?.spotify?.display_name || 'Spotify',
+    label: session.value.user?.spotify || 'Spotify',
     to: '/auth/spotify',
     disabled: Boolean(user.value?.spotify),
     icon: 'i-simple-icons-spotify',
   },
   {
-    label: session.value.user?.google?.email || 'Google',
+    label: session.value.user?.google || 'Google',
     to: '/auth/google',
     disabled: Boolean(user.value?.google),
     icon: 'i-simple-icons-google',
   },
   {
-    label: session.value.user?.twitch?.login || 'Twitch',
+    label: session.value.user?.twitch || 'Twitch',
     to: '/auth/twitch',
     disabled: Boolean(user.value?.twitch),
     icon: 'i-simple-icons-twitch',
   },
   {
-    label: user.value?.auth0?.email || 'Auth0',
+    label: user.value?.auth0 || 'Auth0',
     to: '/auth/auth0',
     disabled: Boolean(user.value?.auth0),
     icon: 'i-simple-icons-auth0',
   },
   {
-    label: user.value?.discord?.username || 'Discord',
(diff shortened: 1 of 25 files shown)


PGLongo:
@Atinux I think that #56 doest not fix this.

```const redirectUrl = getRequestURL(event).href
    if (!code) {
      const scope = config.scope && config.scope.length > 0 ? config.scope : ['User.Read']
      // Redirect to Microsoft Oauth page
      return sendRedirect(
        event,
        withQuery(authorizationURL as string, {
          client_id: config.clientId,
          response_type: 'code',
          redirect_uri: redirectUrl,
          scope: scope.join(' '),
          ...config.authorizationParams
        })
      )
    }
```

The problem was the line 

```const redirectUrl = getRequestURL(event).href' ```

but this line is not affected by #56.

I think that it should be something like:     

`const redirectUrl = config.authorizationParams?.redirect_uri || getRequestURL(event).href`

Do you agree with me?

Atinux:
Actually it's overwritten by the config on https://github.com/Atinux/nuxt-auth-utils/blob/ac9486e634ce12b39048d4016bff651acd082a68/src/runtime/server/lib/oauth/google.ts#L90

Si it should work :)

Issue #47, opened 2024-02-05
Only expose public data part of session


pi0:
Context: H3 sessions are encrypted and only readable by server-side. This can guarantee two things:
- Only server can mutate session so it's data is reliable
- Only server can read/decrypt session so it's data is private

Auth-utils, exposes an edpoint ([`session.get`](https://github.com/Atinux/nuxt-auth-utils/blob/main/src/runtime/server/api/session.get.ts)) that server-side decrypts the session for user. It takes away the second benefit of session encoding which can guarantee data remains secret and private.

While there must be good benefits of this, it is something IMO insecure to do by default and developers might wrongly put sensitive data based on encryption guarantee that will be exposed again.

I would highly recommend (as a breaking change) to only expose `data.public` part of the session.

harlan-zw:
For context, the reason why this came up is that I was trying to store _private_ data into the users sessions. Knowing how sessions work, I figured it would be safe to add it. Afterall it was only data I needed within a Nitro context.

I missed the comments in the code example from the documentation explaining that the actual logic is that nothing attached to the user sessions is safe. I found it later via the jsdocs while debugging some code. In hindsight it is kind of obvious that this is the only way it could work.

So in reality there is no way to attach private data to the users session (without some workarounds?), which seems like a missed opportunity.

 I think:
-`data.public` is a great idea and solves the original issue.
- we need a dedicated docs section explaining how session data is stored / exposed given how sensitive it is

Atinux:
To avoid breaking changes, I am thinking of having:

```ts
setUserSession(event, publicData, privateData)
```

And only expose the `publicData` to the endpoint. I need to think about the types though

Gerbuuun:
Do you already have something in mind? I can have a go at this as I have good use for it.

An easy way is to just have two types and merge them into a single object:
```ts
export interface PublicSessionData {
  user?: User
}

export interface PrivateSessionData {
}

// Like this?
export interface SessionData {
  public?: PublicSessionData
  private?: PrivateSessionData
}
// Or this?
export interface SessionData extends PrivateSessionData {
  public?: PublicSessionData
}
```
and then `/api/session.get`:
```ts
export default eventHandler(async (event) => {
  const session = await requireUserSession(event)

  await sessionHooks.callHookParallel('fetch', session, event)

  return session.public
})
```

---

Or we could do some crazy key manipulation in a way that private properties can be filtered out recursively.
But that does not sound like the way to go.

[...]

PR #65, opened 2024-03-11
feat: private session data

Gerbuuun:
I have added a `public` object to the `SessionData` which still contains the `User` type like before.
The `setUserSession` and `replaceUserSession` have the parameters as described in #47 

The `SessionData` type is no longer exposed in favor of `PublicSessionData` and `PrivateSessionData`.
(only used internally from now on)

I'm not really sure about this solution yet.
What I'm thinking is to let go of the `user` property as a session check and instead just use `public` like `user` is used right now. There won't be a built in computed `user` but this can easily be added by wrapping the `useUserSession`:
```ts
export default function userStore() {
  const { session, clear, fetch, loggedIn } = useUserSession();

  const user = computed(() => session.value.user || null);

  return {
    user,
    session,
    clear,
    fetch,
    loggedIn
  };
}
```
This way you don't have to use `user` and can add anything you want (as long as there is _something_)
Problem is.. `useUserSession` is already taken as composable name

### DIFF
--- a/src/runtime/composables/session.ts
+++ b/src/runtime/composables/session.ts
@@ -1,7 +1,7 @@
 import { useState, computed, useRequestFetch } from '#imports'
-import type { UserSession, UserSessionComposable } from '#auth-utils'
+import type { PublicSessionData, UserSessionComposable } from '#auth-utils'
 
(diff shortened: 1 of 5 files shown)


Gerbuuun:
implemented in #171

PR #71, opened 2024-03-24
fix: narrowed session type passed to `fetch` session hook

Gerbuuun:
The return value of `requireUserSession` is guaranteed to have the `user` component defined.
Previously, the return type was changed to reflect that fact but the session type passed to the `fetch` hook was not.
(It was not immediately clear to me, but the result of `requireUserSession` is passed into the `fetch` hook)
This fixes that and should avoid users to do an unnecessary check when hooking into `fetch`

Previously:
```ts
export default defineNitroPlugin(() => {
  sessionHooks.hook('fetch', (session, event) => {
    // session.user is possibly undefined, so we add a guard
    if (!session.user)
      // Do something

    // Do what you actually wanted to do with 'session.user'
  });
});
```

With this PR:
```ts
export default defineNitroPlugin(() => {
  sessionHooks.hook('fetch', (session, event) => {
    // session.user is defined so you can immediately do things without the need for a check
  });
});
```

Also, `ActiveUserSession` is not exposed because the type is completely defined by `UserSession` and `User`.

### DIFF
--- a/src/runtime/server/utils/session.ts
+++ b/src/runtime/server/utils/session.ts
@@ -4,7 +4,7 @@ import { defu } from 'defu'
 import { createHooks } from 'hookable'
 import { useRuntimeConfig } from '#imports'
 import type { UserSession } from '#auth-utils'
-import { ActiveUserSession } from '../../types/session'
+import type { ActiveUserSession } from '../../types/session'
 
 export interface SessionHooks {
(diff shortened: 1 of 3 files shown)


Atinux (approved):

Atinux:
Thanks!

PR #76, opened 2024-04-07
feat: add opts to requireUserSession for error message and status code customization

dethdkn:
Hello, This PR introduces the ability to customize the status code and error message in the `requireUserSession` method. This feature is particularly useful for applications that are not in English.

I use the returned error message on my apps, and I wish I could translate the `Unauthorized` message to my language


src/runtime/server/utils/session.ts
```diff
@@ -59,13 +59,18 @@ export async function clearUserSession (event: H3Event) {
   return true
 }
 
-export async function requireUserSession(event: H3Event): Promise<UserSessionRequired> {
+/**
+ * Require a user session
+ * @param event
+ * @param opts Options to customize the error message and status code
+ */
+export async function requireUserSession(event: H3Event, opts:{statusCode?: number, message?:string} = {}): Promise<UserSessionRequired> {
```
Atinux, line 67:
```suggestion
export async function requireUserSession(event: H3Event, opts: { statusCode?: number, message?: string } = {}): Promise<UserSessionRequired> {
```

Atinux:
Thank you for the PR @dethdkn 

Don't you want to translate on the application side instead of the API @dethdkn ?

dethdkn:
Hey @Atinux I translate both 😅, Most of my $fetch are like this

```ts
const res = await $fetch('/do-something', { method: 'post', body: body.data })
	.catch((err) => { toast.add({ title: err.data.message, icon: 'i-heroicons-exclamation-triangle', color: 'red' }) })
```

I get the returned api message and show it as a toast
I'm not sure if it's the best way to work with $fetch catch, but, it has been working here since my api have multiple throws.

dethdkn:
Also sorry for the inconsistent spacing on the opts type, I thought eslint would fix that

Atinux:
Happy to fix the conflicts with the new eslint rules? see #79

dethdkn:
Fixed @Atinux 🫡

Atinux (approved):

PR #77, opened 2024-04-10
docs: removed reference to /api in readme

onmax:
Looking at the playground and demos, the code always uses the server route `/auth/PROVIDER`, but the README refers to the `/api/auth/PROVIDER`, which is confusing because the route you create is not inside the API.

I am completely new to this package, so I may be misunderstanding something. 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ const { loggedIn, user, session, clear } = useUserSession()
   </div>
   <div v-else>
     <h1>Not logged in</h1>
-    <a href="/api/auth/github">Login with GitHub</a>
+    <a href="/auth/github">Login with GitHub</a>
   </div>
 </template>
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -190,7 +190,7 @@ We leverage hooks to let you extend the session data with your own data or to lo
 ```ts
 // server/plugins/session.ts
 export default defineNitroPlugin(() => {
-  // Called when the session is fetched during SSR for the Vue composable (/api/_auth/session)
+  // Called when the session is fetched during SSR for the Vue composable (/_auth/session)
```
Atinux, line 193:
```suggestion
  // Called when the session is fetched during SSR for the Vue composable (/api/_auth/session)
```

Atinux (approved):

Atinux:
I kept it for session internal call as it uses /api

PR #83, opened 2024-04-17
fix: avoid duplicate trigger of session fetch hook due to request retry

Diizzayy:
The nitro session fetch hook is likely to be used for implementing the necessary checks to ensure token validity, this might require making additional network requests to the pertinent authentication server.

The current implementation results in the aforementioned checks being made twice whenever the **session cookie is present**, and the **earlier checks have failed** (_returned `throw createError()` within session fetch hook_), hence resulting in unnecessary overhead.

### DIFF
--- a/src/runtime/composables/session.ts
+++ b/src/runtime/composables/session.ts
@@ -19,6 +19,7 @@ async function fetch() {
     headers: {
       Accept: 'text/json',
     },
+    retry: false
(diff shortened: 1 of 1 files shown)


Atinux (approved):

Atinux:
Nice one @Diizzayy ❤️

PR #86, opened 2024-05-03
feat: add fields support to facebook provider

ozancakir:
* Facebook provider was only fetching 'id' and 'name' fields, I have added config.fields support.
* Facebook login redirects with /#_=_ , to avoid ugly and wrong url hash, I have added how to remove it efficiently with nuxt3 server hooks to playground, (https://stackoverflow.com/questions/7131909/facebook-callback-appends-to-return-url)


Atinux:
Thanks @ozancakir 

I moved the server plugin in the core directly if facebook is enabled :)

josef-ngp:
literally working on a new project in need of this and found this PR, thanks every1 ♥️

PR #108, opened 2024-06-23
feat: add paypal as supported oauth provider

Yizack:
For testing if needed, follow these steps:

1. Login in sandbox mode https://developer.paypal.com/dashboard/applications/sandbox
2. Go to "REST API apps" section, select or create an app, and copy the credentials needed for the .env file
3. Click on your app name, scroll down to the "Features" section, and enable "Log in with PayPal"
4. Click on the "Advanced Settings" text and place a "Return URL"
5. Paypal gives an error if you try adding `http://localhost:3000/auth/paypal` so I manually added the `--host` flag on the `dev` script to add my network IP address as a Return URL
6. Scroll down, select the scope email or the ones that you would like to test, and click on save
7. Test the Paypal auth

image

I also added the `sandbox` config which allows using the sandbox or production environment from Paypal, used `import.meta.dev` as the default value to automatically set it true in dev and false in production when not defined

https://github.com/Atinux/nuxt-auth-utils/blob/1ad2a93e4a06d127f9aa22a436f393883a4ea927/src/runtime/server/lib/oauth/paypal.ts#L63-L68


amandesai01 (approved):
LGTM

Atinux:
Sorry for the late review @Yizack 

I am getting this error even though I followed your steps.

CleanShot 2024-07-07 at 16 52 27@2x

Thanks for helping btw <3

Atinux:
Will merge and feel free to continue on this conversation or add another PR to fix later on.

Yizack:
Hey! @Atinux 
> I am getting this error even though I followed your steps.

This error happens to me only when using the default business sandbox account. I am not sure why but try creating another sandbox account here https://developer.paypal.com/dashboard/accounts

Also for testing I realized is better to use `http://127.0.0.1:3000/auth/paypal` as redirect URL than your --host ip

PR #121, opened 2024-07-11
feat: X/Twitter email requirement enhancement

fayazara:
Similar to Github, if config.emailRequired is set, then we make a request to get the users email id from twitter.

### DIFF
--- a/src/runtime/server/lib/oauth/x.ts
+++ b/src/runtime/server/lib/oauth/x.ts
@@ -27,9 +27,15 @@ export interface OAuthXConfig {
    * X OAuth Scope
    * @default []
    * @see https://developer.x.com/en/docs/authentication/oauth-2-0/user-access-token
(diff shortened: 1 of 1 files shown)


src/runtime/server/lib/oauth/x.ts
```diff
@@ -27,10 +27,16 @@ export interface OAuthXConfig {
    * X OAuth Scope
    * @default []
    * @see https://developer.x.com/en/docs/authentication/oauth-2-0/user-access-token
-   * @example [ 'tweet.read','users.read','offline.access ],
+   * @example [ 'tweet.read','users.read','offline.access' ],
```
Atinux, line 30:
```suggestion
   * @example ['tweet.read', 'users.read', 'offline.access'],
```

Atinux (approved):

PR #123, opened 2024-07-12
refactor!: prefer one export per provider

### LINKED ISSUE #49: Prefer Single Exports Over `oauth`
Currently, users will set up their endpoints by accessing one of the `oauth.*`. 

```ts
export const oauth = {
  githubEventHandler,
  spotifyEventHandler,
  googleEventHandler,
  twitchEventHandler,
  auth0EventHandler,
  microsoftEventHandler,
  discordEventHandler,
  battledotnetEventHandler,
  keycloakEventHandler,
  linkedinEventHandler,
  cognitoEventHandler
};
```

This works and the auto-completion is nice but I think it can be improved:
- treeshaking afaik won't work as all providers are within the same const, meaning we will bundle all providers (which will grow)
- `oauth` seems like a likely for collisions in global import namespace

Exporting each provider as its own function should fix tree shaking and namespace problems, i.e `authGoogleEventHandler` where autocompletion should be good.

Barbapapazes:
fix #49

Hello 👋,

Each providers are now their own export.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ export {}
 
 All helpers are exposed from the `oauth` global variable and can be used in your server routes or API routes.
 
-The pattern is `oauth.<provider>EventHandler({ onSuccess, config?, onError? })`, example: `oauth.githubEventHandler`.
+The pattern is `oauth<Provider>EventHandler({ onSuccess, config?, onError? })`, example: `authGithubEventHandler`.
 
 The helper returns an event handler that automatically redirects to the provider authorization page and then calls `onSuccess` or `onError` depending on the result.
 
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -124,7 +124,7 @@ export {}
 
 All helpers are exposed from the `oauth` global variable and can be used in your server routes or API routes.
 
-The pattern is `oauth.<provider>EventHandler({ onSuccess, config?, onError? })`, example: `oauth.githubEventHandler`.
+The pattern is `oauth<Provider>EventHandler({ onSuccess, config?, onError? })`, example: `authGithubEventHandler`.
```
Barbapapazes, line 127:
```suggestion
The pattern is `auth<Provider>EventHandler({ onSuccess, config?, onError? })`, example: `authGithubEventHandler`.
```

PR #130, opened 2024-07-30
fix: always return 200 for session endpoint

### LINKED ISSUE #115: nuxt-auth-utils throws `Unauthorized (/api/_auth/session)` on any 404 page
I added sentry to my project and within days I had hundreds and hundreds of `Unauthorized` errors coming up. 
After investigating I arrived to the conclusion that:

```
nuxt-auth-utils throws `Unauthorized (/api/_auth/session)` on any 404 page
```
So since bots scan for /wp-admin etc I get a ton of these. 
This is a small reproduction (create a nitro plugin):

```js
// server/plugins/server_error_handler.js

export default defineNitroPlugin((nitroApp) => {
    nitroApp.hooks.hook('error', (err, context) => {
        console.error(`${err?.message || "Server Error"} (${context?.event?.path})
    }
})
```

Open any 404 page like `/api/does-not-exists` or even client pages like `/does-not-exists`

The console will show:

```
 ERROR  Unauthorized (/api/_auth/session)
 ```
 
 Not sure why this is happening but it doesn't seem like a normal behavior.

### PR DESCRIPTION
**Pull request by Atinux**
Resolves #115

### DIFF
--- a/src/runtime/server/api/session.get.ts
+++ b/src/runtime/server/api/session.get.ts
@@ -1,10 +1,13 @@
 import { eventHandler } from 'h3'
-import { requireUserSession, sessionHooks } from '../utils/session'
+import { getUserSession, sessionHooks } from '../utils/session'
+import type { UserSessionRequired } from '#auth-utils'
 
 export default eventHandler(async (event) => {
-  const session = await requireUserSession(event)
+  const session = await getUserSession(event)
 
-  await sessionHooks.callHookParallel('fetch', session, event)
+  if (session.user) {
+    await sessionHooks.callHookParallel('fetch', session as UserSessionRequired, event)
+  }
 
   return session
 })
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/runtime/server/api/session.get.ts [hunk 1]
```diff
@@ -1,10 +1,13 @@
...
+import type { UserSessionRequired } from '#auth-utils'
 
 export default eventHandler(async (event) => {
-  const session = await requireUserSession(event)
+  const session = await getUserSession(event)
 
-  await sessionHooks.callHookParallel('fetch', session, event)
+  if (session.user) {
+    await sessionHooks.callHookParallel('fetch', session as UserSessionRequired, event)
+  }
```
katlyn, line 10: Is there a reason for requiring that `user` is defined on the session before running the `fetch` hook? This seems to be a breaking change, compared to the previous implementation. For my use case i would like the `user` property to be populated dynamically by a session hook based on other information in the session, but this change prevents that use case.

PR #156, opened 2024-09-03
feat: add `hashPassword` & `verifyPassword` server utils

### LINKED ISSUE #155: Add `hashPassword()` and `verifyPassword()` server utils
```ts
// import is optional
import { hashPassword, verifyPassword } from '#auth-utils'

const hash = hashPassword('user_password')
// $2a$10$vEdoE9pqBlkJMqWmbk1JWuEYbJoansodgPo7a2GBkXsO8SleGXCr6

if (verifyPassword(existingHash, plainTextValue)) {
  // password is correct
}
```

Demo from bcrypt-edge on https://github.com/atinux/nuxt-bcrypt-edge

The second argument of `hashPassword` is the same as [hashSync](https://github.com/bruceharrison1984/bcrypt-edge/blob/e1b41aa46c221ed5b92a59541701add6bdbfc49d/src/bcrypt-edge.ts#L38)

```ts
const hash = hashPassword('use_password', 12)
```

Barbapapazes:
resolves #155

This PR introduces 2 news utilities that can be used to register and login a user with a password:

- `hashPassword(password)`
- `verifyPassword(hashedPassword, password)`

This PR is still in draft because I need to rework the current `login` route (used with predefined password).

- [x] Move current `login` to another route (with a better name
- [x] Reuse `login` for log in a user after a registration (`register`).

_possible thanks to the work of Atinux (https://github.com/atinux/nuxt-bcrypt-edge)_


carlos-duran:
Thank you so much for this feature!

I implemented my login endpoint in a similar way, but having it out of the box is really great.

In mi case I had to use `await fetch()` (**await**) because I do a navigation and I have an auth middleware in the target page.

```ts
// app/pages/login.vue

//...
const { fetch } = useUserSession()

async function onSubmit(event: FormSubmitEvent<LoginDTO>) {
  $fetch("/api/auth/login", { method: "POST", body: event.data })
    .then(async () => {
      await fetch();
      navigateTo("/");
    })
    .catch((error: FetchError) => {
      toast.add({ title: "Error", description: error?.data.message, color: "red" });
    });
}
```

```ts
// server/api/auth/login.post.ts
import { Admin, LoginDTOSchema } from "@app/schemas";

export default defineEventHandler(async (event) => {
  const body = await readBody(event);
  const { data, error } = LoginDTOSchema.safeParse(body);

  if (error) {
    throw createError({
      statusCode: 400,
      message: "Invalid request body",
      data: error.errors,
    });
  }

  const db = useDB();
  const admin = await db.collection<Admin>("admins").findOne({
    email: data.email,
  });

  if (!admin) {
    throw createError({
      statusCode: 401,
      message: "Invalid email or password",
    });
  }

  const isValid = await Bun.password.verify(data.password, admin.password);

  if (!isValid) {
    throw createError({
      statusCode: 401,
      message: "Invalid email or password",
    });
  }

  const { password: _, ...user } = admin;

  await setUserSession(event, { user });

  return {
    message: "Logged in successfully",
  };
});
```

```ts
// app/middleare/auth.global.ts
export default defineNuxtRouteMiddleware((to) => {
  const { loggedIn } = useUserSession();

  if (to.path !== "/login") {
    if (!loggedIn.value) {
      return navigateTo("/login");
    }
  } else {
    if (loggedIn.value) {
      return navigateTo("/");
    }
  }
});

```

Kudos!

IsraelOrtuno:
I came up with right the same solution too, this is a really great improvement.

atinux:
So, I also added support for `scrypt` which is support natively in CF workers environment: https://github.com/atinux/nuxt-bcrypt-edge/blob/main/server/api/scrypt/hash.post.ts

It does need a [small hack](https://github.com/atinux/nuxt-bcrypt-edge/blob/f398c6993405a9e1d385412e63744c999e4f8396/nuxt.config.ts#L22-L26) though in the meantime (working with Pooya to avoid this)

anuragkumar19:
We should also add a utility for checking compromised passwords.

https://haveibeenpwned.com/API/v3#PwnedPasswords

adrienZ:
What would your recommandations to implement password salting with this ? :)

PR #161, opened 2024-09-05
feat: add gitlab provider

blumgart:
To create a new application for your user:

1. On the left sidebar, select your **avatar**.
2. Select **Edit profile**.
3. On the left sidebar, select **Applications**.
4. Select **Add new application**.
5. Enter a **Name** and **Redirect URI**.
6. Select **OAuth 2 Scopes** as defined in Authorized Applications.
7. In the **Redirect URI**, enter the URL where users are sent after they authorize with GitLab.
8. Select **Save application**. GitLab provides:
-   The OAuth 2 Client ID in the Application ID field.
-   The OAuth 2 Client Secret, accessible by selecting Copy in the Secret field.


src/runtime/server/lib/oauth/gitlab.ts
```diff
@@ -0,0 +1,148 @@
...
+   */
+  clientId?: string
+  /**
+   * GitLab OAuth Client Secret
+   * @default process.env.NUXT_OAUTH_GITLAB_CLIENT_SECRET
+   */
+  clientSecret?: string
+  /**
+   * GitLab OAuth Scope
+   * @default []
```
atinux, line 27:
```suggestion
   * @default ['read_user']
```

src/runtime/server/lib/oauth/gitlab.ts
```diff
@@ -0,0 +1,148 @@
...
+        onError,
+      )
+    }
+
+    const redirectURL = config.redirectURL || getOAuthRedirectURL(event)
+
+    if (!query.code) {
+      config.scope = config.scope || []
+      if (!config.scope.length) {
+        config.scope.push('read_user', 'email')
```
atinux:
```suggestion
        config.scope.push('read_user')
```

PR #162, opened 2024-09-06
feat: add instagram provider

sandros94:
Adding Instagram as new provider.

## To test the provider

1. signup to https://developers.facebook.com
2. create a new app at https://developers.facebook.com/apps
3. while creating the app, at `Use Cases`, pick `Other` (last option)
4. once created, in the `Dashboard`, scroll down to `Add products to your app` and click `set up` for `Instagram Basic Display`
5. once at `Basic display`, you'll see the Instagram App ID and Secret to use as the provider. Alongside, in this page, you'll find:
    - `Valid OAuth Redirect URIs`: I've used Cloudflare's tunnel to quickly test this with https
    - `Deauthorize callback URL` and `Data Deletion Request URL`: just set the domain for now
    - `User Token Generator`: this is important, since you need to invite yourself (or someone else) as the test user

### Add a test user:
1. Under `User Token Generator` click on `Add or Remove Instagram Testers`
2. then click `Add People` button in the top right
3. select `Instagram Tester` at the bottom of that modal and search for the username to add
4. open Instagram's `Apps and Websites` setting page (if you are still on the app user's list there is a convenient link to click
    image
5. go to the `Tester Invites` (all the way to the right) and accept it
6. test the login
    image


atinux (approved):
Thank you!

PR #205, opened 2024-10-05
feat: `clearUserSession` takes config arg as well

### LINKED ISSUE #204: Session cookie not emptied when setting "domain" attribute at creation
Hello there ! 

I created a session with a domain starting with a `dot` so I can get session on subdomains as well.

```ts
      await setUserSession(event, {
        user: user,
        loggedInAt: new Date().toISOString(),
      }, {
        cookie: {
          domain: '.domain.tld',
        },
      })
```

It actually creates a cookie with the `domain: .domain.tld` attribute.  
BUT when I clear the session with `await clearUserSession(event)`, it actually creates a new cookie, with the `domain: domain.tld` attribute, and sets its value blank.

So I'm not disconnected. There's more : if I click "logout" from a subdomain, it creates another empty cookie with `domain: sub.domain.tld`, so the subdomain is logged out, but the others are not.

I see from [h3 code](https://github.com/unjs/h3/blob/7324eeec854eecc37422074ef9f2aec8a5e4a816/src/utils/session.ts#L199)
that `clearSession` can take some `Partial<SessionConfig>` as well and use it when setting the new empty session cookie.

I see from [nuxt-auth-utils code](https://github.com/atinux/nuxt-auth-utils/blob/e0396255b2ebc6b3ef89b9dd62f0eb0fbf345389/src/runtime/server/utils/session.ts#L64) that `clearUserSession` does not take any `config` parameter, making it impossible to reuse the config used in `setUserSession` to pass it to the setCookie called from h3 ?

I'm still a learner, but do you think it would be possible to simply change the `clearUserSession` this way so I can pass the same cookie options I used with `setUserSession`, and the right cookie would be emptied ?

```ts
export async function clearUserSession(event: H3Event, config? Partial<SessionConfig>) {
  const session = await _useSession(event, config)

  await sessionHooks.callHookParallel('clear', session.data, event)
  await session.clear()

  return true
}
```

Again, I'm just trying to start from my problem and reading code from repo to repo, to see how I can fix it, so I'm not sure at all.
Still, if you can confirm my guess is right, it would be an honor to make my first PR to this amazing module.

Thanks A LOT for all your work, it's a real pleasure every day to use the amazing stuff you create :heart:

mathieunicolas:
resolves #204

### DIFF
--- a/src/runtime/server/utils/session.ts
+++ b/src/runtime/server/utils/session.ts
@@ -61,8 +61,8 @@ export async function replaceUserSession(event: H3Event, data: UserSession, conf
  * @param event The Request (h3) event
  * @returns true if the session was cleared
  */
-export async function clearUserSession(event: H3Event) {
-  const session = await _useSession(event)
+export async function clearUserSession(event: H3Event, config?: Partial<SessionConfig>) {
+  const session = await _useSession(event, config)
 
   await sessionHooks.callHookParallel('clear', session.data, event)
   await session.clear()


atinux:
Welcome to the contributors!

PR #212, opened 2024-10-09
fix: tiktok should follow the same config naming convention

navanjr:


atinux:
Well weird enough, TikTop is the only OAuth provider waiting for a `client_key` as parameter...

PR #254, opened 2024-10-23
chore: session max age note

patrick-hofmann:
I've added a note and example since it is not obvious that maxAge of the session is not the same as maxAge of the cookie

### DIFF
--- a/README.md
+++ b/README.md
@@ -589,10 +589,15 @@ export default defineNuxtConfig({
   runtimeConfig: {
     session: {
       maxAge: 60 * 60 * 24 * 7 // 1 week
+      cookie: {
+        maxAge: 60 * 5 // 5 minutes
+      }
     }
   }
 })
 ```
+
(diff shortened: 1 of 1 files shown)


atinux:
Hi @patrick-hofmann 

Actually we use `maxAge` to set the cookie expiration, see https://github.com/unjs/h3/blob/7324eeec854eecc37422074ef9f2aec8a5e4a816/src/utils/session.ts#L141

As the session is stored in the cookie, it expires at the same time the cookie expires.

patrick-hofmann:
Hi @atinux 

you are right, indeed session and cookie lifetime are equal in the current example. Nevertheless this documentation PR also shows why I've created the two PRs 

#255 - cookies shall live as short as possible and lifetime should extend while in usage (this could also be achived with a kind of session rotation)

and

#257 - the session lifetime shall end immediately when the user is logged out.

Best regards!

PR #255, opened 2024-10-23
auto extend session cookie

patrick-hofmann:
With this PR, the session cookie lifetime is extended on each request to the server


atinux:
Thank you, I believe this should be documented instead as I don't want to slow down the requests by adding a global middleware in this module.

An update in the readme on how to easily extend the user session would be great!

PR #257, opened 2024-10-24
add session revocation

patrick-hofmann:
Add possibility to revoke sessions to prevent session stealing from cookies for logged out users

### DIFF
--- a/playground/nuxt.config.ts
+++ b/playground/nuxt.config.ts
@@ -24,5 +24,7 @@ export default defineNuxtConfig({
   },
   auth: {
     webAuthn: true,
+    sessionRevocationStorage: 'revoked-sessions',
+  },
   },
 })
--- a/src/module.ts
+++ b/src/module.ts
@@ -23,6 +23,10 @@ export interface ModuleOptions {
    */
   webAuthn?: boolean
   /**
+   * Storage location for revocation data
+   */
+  sessionRevocationStorage?: string
+  /**
    * Hash options used for password hashing
    */
   hash?: {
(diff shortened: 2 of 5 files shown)


atinux:
Thank you @patrick-hofmann 

But I believe in order to do this we need to properly think of creating a more generic Session Storage so it can be used instead to track active session instead of storing revoked one.

If I am not mistaken I believe @Barbapapazes is working on something like this.

Barbapapazes:
Hello 👋,

Indeed, I'm working on something but I can't give an ETA (I'm making progress but it's bigger than expected).

With a cookie store, you can't revoke a session but you don't need to store the session. It's a compromise compared to a database session.

You can't revoke a cookie session because you're using the user's device as a shop and so the shop isn't centralised. For this reason, it is not possible to get all the sessions of the same user.

If the session is on a stolen device, you'll never be able to revoke a session. With a database session, a user can view all active session and delete them (on the server) so the cookie won't be usable.

patrick-hofmann:
Hi @atinux and @Barbapapazes

Can I help you with your work in progress solution?

Best regards,
Patrick

patrick-hofmann:
Hi @atinux and @Barbapapazes 👋

i've added a PR #265 to provide a server side session and therefore allow revoking sessions by logging out the user. I've used useStorage in order to persist the data since it's less effort also for beginners in contrast to a full database implementation and has less dependencies.

Maybe @Barbapapazes we can merge your progress at some point?

For a complete server side handling of users with different login methods we could think of nuxt-auth-utils being an identity provider itself and federating the login to the different providers. This way sessions of a user with a unique id across providers could be related and managed across devices, but I think this is an even bigger decision/implementation and i'm not sure if this is still the scope of this module or if this should be a separate module (nuxt-auth-idp?)

PR #264, opened 2024-10-29
feat: add laravel passport provider

rhoseno98:
Laravel passport oauth

### DIFF
--- a/playground/app.vue
+++ b/playground/app.vue
@@ -147,6 +147,12 @@ const providers = computed(() =>
       disabled: Boolean(user.value?.polar),
       icon: 'i-iconoir-polar-sh',
     },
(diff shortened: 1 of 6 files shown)


atinux:
From what I see, this seems to be a generic OAuth Provider, wondering if we should just have defineOAuthEventHandler that need to get config directly to work

rhoseno98:
yes, here to minimize the repetitive functions  for access to the laravel passport provider. so that the configuration can be simpler

DiyanDev:
Any plans on merging this, or making a generic OAuth provider available?

halimkun:
> Any plans on merging this, or making a generic OAuth provider available?

+1

PR #265, opened 2024-10-30
feat: add server side session storage

patrick-hofmann:
This PR adds a server side storage option for sessions using `useStorage`.

**The default behavior after this PR is still the same as before (using cookie as storageType which is the default)**

```ts
export default defineNuxtConfig({
  modules: ['nuxt-auth-utils'],
  auth: {
    storageType: 'cookie', // 'memory', 'cache', 'nuxt-session'
  }
})
```

but now the storageType can be `memory`, `cache` or `nuxt-session` as well, where `nuxt-session` is a nitro storage mount point that can be defined in the `nuxt.config.ts`.

The 'non-cookie' storageTypes allow two more configs:

```ts
  auth: {
    sessionInactivityMaxAge: 60 * 60 * 24 * 30, // Session timeout after inactivity (30 days)
    autoExtendSession: true // Extend session on each request
  },
```

The `sessionInactivityMaxAge` is used to determine orphaned sessions that can be deleted with `cleanupOrphanedUserSessions`, which can run in a scheduled nitro task (unfortunately not on the edge) or in a server middleware or a specific route.

The session data is not encrypted to the server which can be discussed as the stored data is obvious at least in transfer to the server side anyway.

What this PR can **not** do, is determining multiple sessions of one user as this would need a server side user id, independent from the auth provider used (since one user could have multiple different providers).


src/runtime/server/utils/session.ts
```diff
@@ -37,7 +55,16 @@ export async function getUserSession(event: H3Event) {
 export async function setUserSession(event: H3Event, data: UserSession, config?: Partial<SessionConfig>) {
   const session = await _useSession(event, config)
 
-  await session.update(defu(data, session.data))
+  const sessionStorage = getSessionStorage()
+  if (sessionStorage) {
+    await sessionStorage.setItem(`nuxt-session:${session.id}`, {
+      ...data,
+      lastAccess: Date.now(),
+    })
```
atinux, line 63: We should get the session too here as this is extending the session data normally using `defu`

src/runtime/server/utils/session.ts
```diff
@@ -37,7 +55,16 @@ export async function getUserSession(event: H3Event) {
...
+  }
+  else {
+    await session.update(data)
```
atinux, line 66:
```suggestion
    await session.update(defu(data, session.data))
```

src/runtime/server/utils/session.ts
patrick-hofmann: I've now extended the session using defu according to your comment https://github.com/atinux/nuxt-auth-utils/pull/265/commits/f8ce2ffdc7a274ba942de8e685d01d02ca8a01c4

README.md
```diff
@@ -514,6 +514,22 @@ export default defineNitroPlugin(() => {
...
+Example middleware:
+
+```ts
+// server/middleware/cookie-lifetime-extend.ts
+export default defineEventHandler((event) => {
+  const session = await getUserSession(event)
+  if (session && Object.keys(session).length > 0) {
+    await setUserSession(event, session)
+  }
+}
```
robinsimonklein, line 530: `)` is missing at the end here ;)

README.md
```diff
@@ -514,6 +514,22 @@ export default defineNitroPlugin(() => {
...
 
+### Extending Cookie Lifetime
+
+When using cookie mode, only the session data itself can be enriched through the session hooks. If you want to extend the lifetime of the cookie itself, you could use a middleware that refreshes the cookie by re-setting the session data. (At this point you can also add any other custom data to your cookies)
+
+Example middleware:
+
+```ts
+// server/middleware/cookie-lifetime-extend.ts
+export default defineEventHandler((event) => {
```
robinsimonklein, line 525:
```suggestion
export default defineEventHandler(async (event) => {
```

patrick-hofmann: thanks for your help! @robinsimonklein

patrick-hofmann:
No problem, take the time you need

thewebartisan7:
Any update on this?

atinux:
I did not forget, was running low on bandwidth to properly review it, but it's in my list!

RobertoMike:
Hi, when will this be implemented?
We can help in some way to accelerate the process? 
I would like to have these functionalities.

RedStar071:
Any update?

J0Z14H:
Hey @atinux are server-side sessions for nuxt-auth-utils still on the roadmap?

Issue #272, opened 2024-11-09
How do we link a passkey to an existing user? Since there's only register and authenticate methods


fayazara:
I see that we have `defineWebAuthnRegisterEventHandler` but this is always expecting a userName and when we want to just link an credential to a existing user, its not possible. 

```js
export default defineWebAuthnRegisterEventHandler({
  async onSuccess(event, { credential, passkeyName }) {
    const { user } = await requireUserSession(event);
    // we have the users email here
    return user;
  },
});
```

The above endpoint breaks - it throws 

```json
{
    "url": "/api/auth/link-passkey",
    "statusCode": 400,
    "statusMessage": "",
    "message": "Invalid request, missing userName or verify property",
    "stack": "<pre><span class=\"stack internal\">at createError (./node_modules/.pnpm/h3@1.13.0/node_modules/h3/dist/index.mjs:78:15)</span>\n<span class=\"stack internal\">at Object.handler (./node_modules/.pnpm/nuxt-auth-utils@0.5.2_@simplewebauthn+browser@11.0.0_@simplewebauthn+server@11.0.0_encoding@0_53tvg2bimanm6byvroxwbmjc6q/node_modules/nuxt-auth-utils/dist/runtime/server/lib/webauthn/register.js:20:13)</span>\n<span class=\"stack internal\">at process.processTicksAndRejections (node:internal/process/task_queues:105:5)</span>\n<span class=\"stack internal\">at async ./node_modules/.pnpm/h3@1.13.0/node_modules/h3/dist/index.mjs:1978:19</span>\n<span class=\"stack internal\">at async Object.callAsync (./node_modules/.pnpm/unctx@2.3.1_webpack-sources@3.2.3/node_modules/unctx/dist/index.mjs:72:16)</span>\n<span class=\"stack internal\">at async Object.callAsync (./node_modules/.pnpm/unctx@2.3.1_webpack-sources@3.2.3/node_modules/unctx/dist/index.mjs:72:16)</span>\n<span class=\"stack internal\">at async Server.toNodeHandle (./node_modules/.pnpm/h3@1.13.0/node_modules/h3/dist/index.mjs:2270:7)</span></pre>"
}
```

@Gerbuuun is this handled somewhere that I am missing in the docs or can we not do this currently? 

I can contribute, just want to make sure if this is not implemented already

fayazara:
I was thinking if something like below would make sense 

[...]

PR #275, opened 2024-11-10
feat(webauthn): allow userName in register handler options

fayazara:
Allow passing userName directly to defineWebAuthnRegisterEventHandler to support linking credentials to existing users without requiring userName in request body.

No breaking change as it maintains backward compatibility while adding new functionality.

How it works?

```ts
export default defineWebAuthnRegisterEventHandler({
  userName: 'user@example.com',
  async onSuccess(event, { credential }) {
    // your db actions or whatever you'd like to do with the credential
    return credential
  },
})
```


fayazara:
@atinux @Gerbuuun I have made a small enhancement to the register event handler, please take a look

Issue ref: https://github.com/atinux/nuxt-auth-utils/issues/272

Gerbuuun:
I think it is better to just do the check inside the handler already? That way the dev doesn't need to implement it themselves
```ts
let user: T;
try {
  // If user is logged in, use the current user's username
  const { user } = await requireUserSession(event);
  user = { userName: user.email };
}
catch {
  // Else do the currently implemented checks
  if (!body.user?.userName)
    throw createError({
      message: 'Invalid request, missing userName or verify property',
      statusCode: 400,
    })

  user = body.user
  if (validateUser) {
    user = await validateUserData(body.user, validateUser)
  }
}

// rest of registration event handler..
```

fayazara:
> Other idea: Make
> 
> ```ts
> if (body.verify === undefined || !body.user?.userName)
>   throw createError({
>     message: 'Invalid request, missing userName or verify property',
>     statusCode: 400,
>   })
> ```
> 
> the default `validateUser` function and don't force any shape to the request body?

Didn't quite get this, How would a said api method look like?

```
export default defineWebAuthnRegisterEventHandler({
  async onSuccess(event, { credential }) {
    const { user } = await requireUserSession(event);
    // we have the users email here
    return user;
  },
});
```


Another ideas was to have a separate method for this called `defineWebAuthnConnectEventHandler` - but feels unnecessary

fayazara:
@Gerbuuun Updated this to check if the user is already logged in, this way is a user is already logged in, then we pick the `userName` from the session

Issue #267, opened 2024-10-31
Add ATProto/Bluesky as provider


AdrianoCahete:
Since you have most of the social media (Facebook, Instagram and X) sites, do you plan to integrate to [atproto](https://github.com/bluesky-social/atproto) (bluesky protocol)?

atinux:
Well I don't have time to do it myself but if someone want to open a PR I am happy to review it!

atinux:
Adding this as it could help:

https://github.com/pilcrowonpaper/atproto-oauth-example

noook:
I'll be happy to work on it. 😄

atinux:
Yeah it seems to something, good luck @noook 😄

noook:
Actually got something, I'm handling edge cases 😁

PR #281, opened 2024-11-13
 feat: add bluesky as a provider

noook:
Resolves: https://github.com/atinux/nuxt-auth-utils/issues/267

This PR adds Bluesky as a provider.

This provider requires the user to install extra dependencies to properly handle authorization, because of the way Bluesky works.

In order to begin the authorization process, we first need to know the user handle. This is required because we need to know against which instance of Bluesky we need to verify the user. Doing all the verifications manually require a lot of steps and adds complexity, so we use the `atproto` API instead.


atinux:
Damn you are on 🔥 

Regarding the scopes, I think we can anyway remove the duplicate because such a change on `defu` can have strong side effects, might be easier to do it ourself after the merge.

src/runtime/server/lib/oauth/bluesky.ts
```diff
@@ -0,0 +1,201 @@
...
+  return eventHandler(async (event: H3Event) => {
+    config = defu(config, useRuntimeConfig(event).oauth?.bluesky) as OAuthBlueskyConfig
+
+    config.scope ||= []
+    if (!config.scope.includes('atproto')) {
+      // atproto is a required scope
+      config.scope.push('atproto')
+    }
+
+    const scope = config.scope.join(' ')
```
atinux, line 60:
```suggestion
    const scope = [...new Set(config.scope)].scope.join(' ')
```

noook:
I'm not sure I'm satisfied regarding the current implementation. Basically Bluesky is just a provider using atproto underneath, and we could have sooner or later another atproto provider.

I think I'm facing the same issues that come with the complexity of providing a generic OIDC provider, but with the additional build time constraints on top of that (exposing the discovery document)

So the configuration would be split into two parts:
- `atproto` -> next to `oauth` and `webauthn` configuration. Probably a boolean, so we can check that the peer dependencies are installed.
- `oauth.${atprotoProvider}` -> Should implement an interface such that matches the "Client ID Metadata Document" section on [this document](https://atproto.com/specs/oauth#clients) that seems to be a common base for future atproto providers

I guess the new challenge here is to provide the dynamic metadata handler

src/runtime/server/lib/oauth/bluesky.ts
noook: I don't think this is the way, as of now it's currently causing a memory leak as every request will add the configured scope to the list of scopes. I should patch the merge strategy of `defu` instead

atinux:
Please ping me when good to merge, amazing work you are doing @noook ❤️

noook:
@atinux It should be good. I'm having issue when I need to type `useRuntimeConfig(event).oauth[provider]`

```
Element implicitly has an 'any' type because expression of type 'OAuthProvider' can't be used to index type '{ github: { clientId: string; clientSecret: string; redirectURL: string; }; gitlab: { clientId: string; clientSecret: string; redirectURL: string; }; spotify: { clientId: string; clientSecret: string; redirectURL: string; }; ... 23 more ...; authentik: { ...; }; }'.
```

I understand the error, it makes sense, but I'm not sure how I can do this properly

atinux:
I just noticed that you use `useStorage` which can be tricky as it needs to be configured in serverless environment in order to properly work. One solution could be to use cookies to store the state, see https://github.com/pilcrowonpaper/atproto-oauth-example/blob/2ebfd7836b5ce9921df31e66fb332ee8ae8d5823/src/pages/login/callback.ts#L15-L17

noook:
Sorry, been away for a long time @atinux. I updated the code so that it uses cookies for storing state and session. I think we could delete the session once it's read, as it's not longer used after authentication, wdyt ?

atinux:
>  I think we could delete the session once it's read, as it's not longer used after authentication, wdyt ?

Agree! Happy to do it and fix the conflicts? 🙏 

Don't apologize, it's open source, no hurry at all :D

src/module.ts
```diff
@@ -231,6 +240,35 @@ export default defineNuxtModule<ModuleOptions>({
       clientSecret: '',
       redirectURL: '',
     })
+
+    // Atproto OAuth
+    for (const provider of atprotoProviders) {
+      runtimeConfig.oauth[provider] = defu(runtimeConfig.oauth[provider], atprotoProviderDefaultClientMetadata)
+    }
```
noook, line 247: Leaving this in the `if` block here under causes `runtimeConfig.oauth[<atprotoProvider>` not to be defined, and thus causing a type error.

Issue #280, opened 2024-11-13
OAuth Provider for workos


brianacdev:
https://workos.com/docs/reference/user-management/authentication

I am using WorkOS for auth and have implemented it manually for my project but figured I'd add it as a provider so that I can use it with nuxt-auth-utls

PR #283, opened 2024-11-14
feat: add workos oauth provider

brianacdev:
Resolves: #280 

Adding the oauth provider for WorkOS.


atinux:
Is there a way for me to test it locally & for free @brianacdev ?

Thank you for the PR btw!

brianacdev:
You can setup an account for free. I was able to successfully authenticate to my service. I could also dm you the info for my staging/test environment if you'd prefer that

PR #298, opened 2024-12-03
fix(instagram): oauth provider

### LINKED ISSUE #193: new Instagram API
So, when I implemented Instagram as a provider I wasn't aware of an [API change that was coming](https://developers.facebook.com/blog/post/2024/09/04/update-on-instagram-basic-display-api/?locale=en_US), in particular:

> Starting on December 4th, 2024, [Instagram Basic Display API](https://developers.facebook.com/docs/instagram-basic-display-api/) will no longer be available.
>
> [...]
> - After December 4th, 2024, there will no longer be a set of Instagram APIs for consumer developer apps
>
> [...]
> - Businesses and creators can leverage this new solution or use Instagram API with Facebook Login for their respective business use cases.

On one side this basically kills the project I was working on (🙄), but on the other side I need to update the implementation. So that if someone needs the "Instagram Professional" part they could still use it.

I didn't check what about Facebook's APIs, yet.

@atinux I'll be happy to track, study and work on this in the upcoming weeks. So if anything related comes up feel free to tag me.

sandros94:
Resolves #193 

This both upgrades the Instagram provider to use the new login system (only business and creator accounts are allowed), as well as adds some typing and better error data.

The two most important upstream documentation are:
1. [here](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login/create-a-meta-app-with-instagram) for creating a meta app.
2. [here](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login/get-started) for details on what fields to use for retrieving specific user's data
3. [here](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login/business-login) for more advanced scoped access to authorize for things like messages and content management.

### Note
Currently I left as default my original user's fields:
https://github.com/atinux/nuxt-auth-utils/blob/90d0d180f8d116581d8b095882726efe3ad7db2b/src/runtime/server/lib/oauth/instagram.ts#L122
But there is a not-so-clear difference between `id` and `user_id`:
- `id` refers to the user's id within your current Meta application.
- `user_id` refers to the actual Instagram's unique user id.

@atinux WDYT? Should I change the default to be `user_id` or leave `id` as a general purpose?

### DIFF
--- a/src/runtime/server/lib/oauth/instagram.ts
+++ b/src/runtime/server/lib/oauth/instagram.ts
@@ -19,23 +19,23 @@ export interface OAuthInstagramConfig {
   clientSecret?: string
   /**
    * Instagram OAuth Scope
-   * @default [ 'user_profile' ]
+   * @required [ 'business_basic' ]
    * @see https://developers.facebook.com/docs/instagram-basic-display-api/overview#permissions
-   * @example [ 'user_profile', 'user_media' ],
+   * @example [ 'business_basic', 'business_manage_messages' ],
    */
-  scope?: string[]
+  scope?: ('business_basic' | 'business_content_publish' | 'business_manage_comments' | 'business_manage_messages')[]
 
   /**
    * Instagram OAuth User Fields
    * @default [ 'id', 'username'],
    * @see https://developers.facebook.com/docs/instagram-basic-display-api/reference/user#fields
-   * @example [ 'id', 'username', 'account_type', 'media_count' ],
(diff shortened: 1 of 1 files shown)


atinux:
Let's go with `id` and see if people asks for `user_id`

Thank you for working on it @sandros94 !

sandros94:
> Let's go with `id` and see if people asks for `user_id`

Sounds good

> Thank you for working on it @sandros94 !

With pleasure!

Issue #194, opened 2024-09-30
make requestAccessToken and other utils importable


acidjazz:
I've been adding my own oauth providers [apple](https://github.com/fumeapp/bio/pull/40) , would be helpful to be able to re-use this and other utils.

atinux:
Why not opening a PR to add support for Apple login?

This will also resolve #187

acidjazz:
> Why not opening a PR to add support for Apple login?
> 
> This will also resolve #187

I can look into it but apple does not follow oauth2 standards which a lot of people are not happy about.  Specifically you only get the user information the very 1st time they login with apple, after that you just get their email via the token payload

sans1994:
> > Why not opening a PR to add support for Apple login?
> > This will also resolve #187
> 
> I can look into it but apple does not follow oauth2 standards which a lot of people are not happy about. Specifically you only get the user information the very 1st time they login with apple, after that you just get their email via the token payload

@acidjazz In fact, in certain cases, this is entirely sufficient. For example, the project I am currently working on receives data from providers only during the initial identification phase. That is, when the user logs in for the first time, I send data from the provider to the backend for identification in our system (our API returns an existing user if the email matches, or a new one is created).

So, we do not request data from the provider again, and it is only needed for identification (during login), while all further interaction happens with our backend.

acidjazz:
> > > Why not opening a PR to add support for Apple login?
> > > This will also resolve #187
> > 
> > 
> > I can look into it but apple does not follow oauth2 standards which a lot of people are not happy about. Specifically you only get the user information the very 1st time they login with apple, after that you just get their email via the token payload
> 
> @acidjazz In fact, in certain cases, this is entirely sufficient. For example, the project I am currently working on receives data from providers only during the initial identification phase. That is, when the user logs in for the first time, I send data from the provider to the backend for identification in our system (our API returns an existing user if the email matches, or a new one is created).
> 
> So, we do not request data from the provider again, and it is only needed for identification (during login), while all further interaction happens with our backend.

What about cases where the data from that provider has changed? say name details/avatar/etc ?

sans1994:
> What about cases where the data from that provider has changed? say name details/avatar/etc ?


[...]

Issue #187, opened 2024-09-23
Apple Provider


sans1994:
Hi!)  Are there plans to integrate Apple Provider?

atinux:
No direct plan but if someone open a PR I am happy to merge it!

sans1994:
@atinux Thank you for the feedback! I’ll probably try to implement this on my own and might create a PR, but I'm not sure if I’ll do it or if it will work out.

It would be great if someone could implement this right away. 

P.S. The library is excellent! For a typical project with its own backend, it’s exactly what’s needed. 
Thanks for your work!

atinux:
Checkout #194 as @acidjazz implemented Apple provider on its project!

PR #328, opened 2025-01-26
feat: add apple provider

GreenmeisterDavid:
As I needed the Apple provider in my own project, I figured I'd try and contribute to this awesome package and ecosystem. Feedback/changes to this PR welcome! 💚

## Some things to note: 
- Apple will **only** provide user details the first time a user logs in. Thus, `defineOAuthAppleEventHandler` will only return the `user` once for each user. The `tokens` are always returned (and include a unique `sub` for each user which can be used to match with user info stored before).
- I added [Jose](https://github.com/panva/jose) as a dependency to sign and verify JWTs, as it has no dependencies and works in all environments.
- Apple returns a form POST instead of a query in the URL after user login. Therefore, implementations of `defineOAuthAppleEventHandler`, like in [`/playground/server/routes/auth/apple.ts`](./playground/server/routes/auth/apple.ts), **should NOT** have the `.get.ts` suffix, as the handler will handle both `POST` and `GET` requests.  

## How to test:  
1. In [Apple's developer platform](https://developer.apple.com/account/resources/identifiers/list), create:  
   - An **AppID**,  
   - A **ServiceID**,  
   - A **KeyID and Key**   (set full key in env, linebreaks must be replaced with `\n`)
   - And you need to copy your TeamID from here
   
   *(I found this [guide](https://medium.com/@artyomefremov/add-sign-in-with-apple-button-to-your-website-today-part-1-12ed1444623a) helpful.)*

2. Add these credentials to your `.env` file and use the `defineOAuthAppleEventHandler`. Remember not to use `.get.ts`-suffix to the auth route, as Apple will POST. Just naming it `apple.ts` will work).

> **For testing/development:**  
> I found it useful to temporarily set  
> `https://redirectmeto.com/http://localhost:3000/auth/apple`  
> as a custom redirect URL, as Apple doesn't accept `localhost` directly.  
>  
> - If you do this, make sure to temporarily add it to your Apple Service ID redirect URLs.

## Credits:  
Code inspired by [acidjazz](https://github.com/acidjazz)'s custom solution mentioned in [#194](https://github.com/atinux/nuxt-auth-utils/issues/194).

---

Closes [#187](https://github.com/atinux/nuxt-auth-utils/issues/187)


acidjazz:
nice work!

atinux:
Thank you so much ❤️

PR #335, opened 2025-02-03
feat: add Gitea Oauth Provider

justserdar:
I needed a Gitea provider for Nuxt Auth Utils, but couldn't find it, so I made it few days ago.
Came back to PR the provider, it's nearly identical to GitLab.

If I'm missing something, let me know.
I moved over to GitLab because of Gitea not being a valid giget source in Nuxt as of now.


ManUtopiK:
We can almost use Gitlab oauth to use Gitea.
In nuxt.config.ts :
```js
runtimeConfig: {
    oauth: {
      gitlab: {
        authorizationURL: 'https://{{gitea_url}}/login/oauth/authorize',
        clientId: 'xxxx',
        clientSecret: 'xxxx',
        tokenURL: 'https://{{gitea_url}}/login/oauth/access_token',
      }
    }
  }
```

It works until server meet https://github.com/atinux/nuxt-auth-utils/blob/42a2a7a56649bcc5e2637b67bd97edffcbb60afa/src/runtime/server/lib/oauth/gitlab.ts#L146
We just need to use `v1` instead of `v4` !

justserdar:
Nice to know! Thanks for letting me know. The Gitea provider is essentially the Gitlab provider but just preset provider via this PR.

Sadly gitea doesn't work with giget in Nuxt yet. Hope we can get that neatly upgraded too going forward!

atinux:
Thank you ❤️

PR #336, opened 2025-02-04
feat(composable): add `openInPopup(route, { width, height })`

ManUtopiK:
[Kooha-2025-02-04-03-38-18.webm](https://github.com/user-attachments/assets/6879faf1-c06f-4e66-baaa-c8c7746f002b)


src/runtime/app/composables/session.ts
```diff
@@ -2,6 +2,17 @@ import { appendResponseHeader } from 'h3'
...
+  github: {
+    width: 960,
+    height: 600,
+  },
+  gitlab: {
+    width: 1100,
+    height: 650,
+  },
+}
+
```
atinux, line 15: I don't think we should have the providers with fixed width/height

src/runtime/app/composables/session.ts
```diff
@@ -38,12 +49,48 @@ export function useUserSession(): UserSessionComposable {
     }
   }
 
+  const isInPopup = useCookie('temp-nuxt-auth-utils-popup')
+
+  const openInPopup = (provider, route) => {
```
atinux, line 54: I would rather have `onpenInPopup('/auth/github', { width: 690, height: 600 })` with default values for both width & height so we could have `openInPopup('/auth/github')`

src/runtime/app/composables/session.ts
```diff
@@ -38,12 +49,48 @@ export function useUserSession(): UserSessionComposable {
...
+
+    window.open(
+      route,
+      'nuxt-auth-utils-popup',
+      `width=${conf.width}, height=${conf.height}, top=${top}, left=${left}, toolbar=no, location=no, directories=no, status=no, menubar=no, scrollbars=no, resizable=no, copyhistory=no`,
+    )
+
+    watch(isInPopup, () => {
+      fetch()
+    })
```
atinux, line 76:
```suggestion
    watch(isInPopup, fetch)
```

atinux:
Thank you for the ideas, I added somes notes :)

src/runtime/app/composables/session.ts
atinux: As websites can evolve, I don't want to myself maintain the list of with & height for each provider, also, this would make the app bundle size bigger to store all of them when in reality you may use only one of two

ManUtopiK: You're right. We can just provide the best width & height for each provider in the readme.md

PR #352, opened 2025-02-12
feat: add u.id provider

ekoeryanto:
U.ID is unique because it provide legit National ID Card Number


alexandre-hallaine:
Might be worth checking if `module.ts`, `oauth-config.ts`, and `README.md` need updates as well.

Issue #109, opened 2024-06-24
Add PKCE functionality to support authentication in SPAs


amandesai01:
I am very well aware of the efforts taken in #12, however, I do not agree with the implementation.

PKCEs are a way for SPAs to authenticate securely. They are protected by `redirect_uri`s and use SHA-256 checksum to prevent XSS attacks.
They are pretty well adopted and almost all decent auth providers allow `PKCE`  for authentication.

Therefore, implementing PKCE for server side code doesn't make sense. Also, PR is not being actively worked on.

**Implementation Proposal**:
A Nuxt plugin for providers which runs client side and performs PKCE logic.

Ref: https://github.com/bitinflow/nuxt-oauth/blob/main/src/runtime/plugin.ts

I am open to PR if we are going to merge it.

Atinux:
Thanks for the issue, this module is focused on Nuxt applications running with a server so far (and it's already quite complex).

I like the idea of the plugin but yours is working only for one specific OAuth endpoint right? We will need to have the logic for each provider in this case.

amandesai01:
Yes, my proposal is to have similar system like in backend. Implement general PKCE and let it be customisable.

shgnplaatjies:
> Yes, my proposal is to have similar system like in backend. Implement general PKCE and let it be customisable.

Have you implemented a workaround?  I'm trying to integrate this into a SPA as well and bumped into a `AADSTS9002325: Proof Key for Code Exchange is required for cross-origin authorization code redemption.` error from Microsoft Entra ID.

itpropro:
> Thanks for the issue, this module is focused on Nuxt applications running with a server so far (and it's already quite complex).
> 
> I like the idea of the plugin but yours is working only for one specific OAuth endpoint right? We will need to have the logic for each provider in this case.

PKCE should also be used on a confidential client flow (auth code grant with a client secret) and the implementation logic should not differ between server side client side, so the utility can be universal.
The steps would be
1. Generate and store a PkceVerifier in the current authorization flow session
2. Generate a PkceCodeChallenge based on the PkceVerifier and attach it to the login request in the `code_challenge` field
3. When sending the callback token request to get a access (and/or id) token, add the PkceVerifier in the code_verifier field
4. Clear the temporary auth session used during oauth flow and use the user session or clear the PKCE fields

Hebilicious:
My 2 cents on this : We should split this package/module into 2, one for apps with servers and one for SPAs/mobile apps.
[...]

PR #362, opened 2025-02-28
feat: add azure b2c oauth provider

n-rowe:
Adds support for Azure B2C oauth.

B2C Requires PKCE and recommends state so these have also been added.
I believe there are discussions relating to PKCE in #109
There may be overlapping and it might be best for me to wait until PKCE is sorted across the entire project before adding b2c.


src/runtime/server/lib/oauth/azureb2c.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    const redirectURL = config.redirectURL || getOAuthRedirectURL(event)
+
+    // guarantee uniqueness of the scope
+    config.scope = config.scope && config.scope.length > 0 ? config.scope : ['openid']
+    config.scope = [...new Set(config.scope)]
+
+    // Create pkce verifier
+    const verifier = await handlePkceVerifier(event)
+    const state = await handleState(event)
```
atinux, line 85: I think this should be named `handlePckeState`

philipimperato:
What confuses me about this implementation is that it assumes an `access_token` from the /token POST request. That requires either a custom API or custom Policies in B2C. It doesn't offer any other way of integrating with the graph.microsoft.com/me api. I must be missing something?

n-rowe:
> What confuses me about this implementation is that it assumes an `access_token` from the /token POST request. That requires either a custom API or custom Policies in B2C. It doesn't offer any other way of integrating with the graph.microsoft.com/me api. I must be missing something?

Hi! I'm not quite sure what you mean.
The `/token` endpoint is an oauth implementation and the B2C docs mention the `access_token` being readily available.
You use `access_tokens` in OAuth to get access to APIs (such as Microsoft graph).

Are you having problems with this provider in your project? 

Some resources:
https://learn.microsoft.com/en-us/azure/active-directory-b2c/protocols-overview
https://learn.microsoft.com/en-us/azure/active-directory-b2c/tokens-overview#token-types

philipimperato:
What if I would only use it for authn? Am I forced to great a scope that exposes an API? I'm also readying that the access token from the /token endpoint can't be used (or is a deprecated) way of authorizing the Microsoft Graph API.

n-rowe:
> What if I would only use it for authn

I based this implementation on the existing oauth providers, which all seem to use a userinfo endpoint.

> I'm also readying that the access token from the /token endpoint can't be used (or is a deprecated) way of authorizing the Microsoft Graph API.

Do you have a link to this? I'd be happy to update the implementation if it's not correct as per microsoft standards.

I believe B2C is on the path to deprecation regardless and Microsoft are trying to push people to use Entra External Identity instead

philipimperato:
This response here:
https://stackoverflow.com/questions/63589005/use-access-token-for-graph-api-in-aad-b2c?utm_source=chatgpt.com

Appears to point to the access_token only being used for `/me` if it's `AD`, not `AD B2C` - though, I can't seem to find it in the link he provided. 

I already have the user info with the `id_token`, I don't need to access the `/me` url, can I send in an `authnOnly` or `skipAccessToken` type param?

n-rowe:
> This response here: https://stackoverflow.com/questions/63589005/use-access-token-for-graph-api-in-aad-b2c?utm_source=chatgpt.com
> 
> Appears to point to the access_token only being used for `/me` if it's `AD`, not `AD B2C` - though, I can't seem to find it in the link he provided.
> 
> I already have the user info with the `id_token`, I don't need to access the `/me` url, can I send in an `authnOnly` or `skipAccessToken` type param?

Yes that seems to make sense, it looks like the graph endpoint only works if your B2C has specific scopes setup for it.
I personally use a custom user_info endpoint.

I do think these providers should support not using the user_info endpoint if an id_token is provided, I can look into making a PR for this to follow up the changes from this one unless you want to take a crack at it

philipimperato:
I can create a PR if you're okay with a `skipUserInfo` param avoiding everything after we get the `id_token`?

philipimperato:
https://nuxtoidc.cloud/configuration offers are `skipAccessTokenParsing`

PR #370, opened 2025-03-10
feat: add logto oauth connector

loickal:
Adds [Logto](https://logto.io) as a oAuth Connector


loickal:
Let me know if I have to edit something.
A quick merge would be cool so that I can start finish building my app :)

alexandre-hallaine:
Adding my support for this PR. I think having the Logto connector directly within `nuxt-auth-utils` is the right move. It centralizes our authentication tools, which is much more efficient than needing separate modules for each provider.

PR #373, opened 2025-03-13
fix(oauth, apple): return tokens in response

oritwoen:
The current implementation of the Apple return is a completely different from other providers (e.g. Google, Facebook).

In the current implementation, tokens are not returned as tokens but payload, which is very problematic when you want to use tokens and the same flow as for the rest of the providers, e.g. sending access_token to an additional backend.

Google returns:
- { tokens: { access_token}

Facebook returns:
- { tokens: { access_token }

the rest of the providers in the repository are similar, and apple returns the decoded jwt as tokens :/

### DIFF
--- a/playground/server/routes/auth/apple.ts
+++ b/playground/server/routes/auth/apple.ts
@@ -9,7 +9,7 @@ export default defineOAuthAppleEventHandler({
         apple: userToSet,
       },
       secure: {
-        ...tokens
+        ...tokens,
       },
       loggedInAt: Date.now(),
     })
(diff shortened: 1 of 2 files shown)


atinux:
Thank you for the PR @oritwoen 

What I am afraid about is that this change is a breaking change and can only be applied to a major version as you changed the signature of the `tokens`

atinux (approved):

oritwoen:
Thanks @atinux! But I was also surprised that no one noticed it, maybe as you write no one has used it yet 😅 

Anyway, if you want to do it as a breaking and another major, no problem for me, for my application I made a patch using pnpm until there is a release for it.

PR #376, opened 2025-03-17
Introduce LiveChat OAuth

mariosant:
This PR introduces LiveChat OAuth.
The implementation is following LiveChat documentation as found [here](https://platform.text.com/docs/authorization/authorization-in-practice#authorization-in-practice).
This has been manually tested through the playground. Let me know if there's anything more to that!


src/runtime/server/lib/oauth/livechat.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    const query = getQuery<{ code?: string }>(event)
+    const redirectURL = config.redirectURL || getOAuthRedirectURL(event)
+
+    const scope = [...new Set([...config.scope!, 'accounts--my:ro'])].join(' ')
+
+    if (!query.code) {
+      config.scope ??= []
+
```
mariosant, line 129:
```suggestion
```

mariosant:
Thanks for the quick review @atinux !

PR #379, opened 2025-03-20
Explicit Return Type on getUserSession

Tmmcmasters:
This should fix the typescript issue where some users, including myself, are not seeing the extended properties of the interface UserSession. This should operate as normal.


Tmmcmasters:
This is a very simple PR. I hope that we can get this approved rapidly.

Tmmcmasters:
I was able to pull this down and utilize this into a broken project. It fixes the type errors that we were receiving

sandros94:
IMO it should also be added to `setUserSession` and `replaceUserSession`, @atinux WDYT?

Tmmcmasters:
> IMO it should also be added to `setUserSession` and `replaceUserSession`, omitting `id`.
> 
> @atinux WDYT?

I can update it if that is what is desired

Tmmcmasters:
@atinux @sandros94 Just wanted to poke a reminder to please look at this. This has been waiting for approval for two weeks

atinux:
Sorry for the delay, actually it should be working without having to explicitly set the types, see working example:

CleanShot 2025-04-01 at 12 22 00@2x

Don't you have this in your project?

Tmmcmasters:
> Sorry for the delay, actually it should be working without having to explicitly set the types, see working example:
> 
> CleanShot 2025-04-01 at 12 22 00@2x
> 
> Don't you have this in your project?

The problem that seems to be arising is that on larger projects, the Typescript implicit typing seems to failing as seen here:
image

This is despite the UserSession having these properties assigned in the type extensions:
image

Whenever I pulled this down and published this into our own npm repo, the explicit typing did rectify the issue. Other users also seem to be having this issue as well where not having explicit typing causes problems.

atinux:
In that case I believe we should return explicit types for all the exported utils, would you be to do that?

Tmmcmasters:
> In that case I believe we should return explicit types for all the exported utils, would you be to do that?

Let me take a look and see the scope of this. I might be able to do something like this

Tmmcmasters:
@atinux I updated all server exported functions related to session. Do you want me to update other things?

Tmmcmasters:
> In that case I believe we should return explicit types for all the exported utils, would you be to do that?

Removed the comment to hide some code....

sandros94:
> Sorry for the delay, actually it should be working without having to explicitly set the types, see working example:
> 
> [...]
> 
> Don't you have this in your project?

I might be wrong but I'm noticing something like a racing condition depending on the ts server version and OS used. And since the local project might be declaring their own `User`, `UserSession` this becomes even more evident on some platform like windows

atinux:
Could you try with:
```bash
npm i https://pkg.pr.new/atinux/nuxt-auth-utils@379
```

And confirm it fixes your issue?

Tmmcmasters:
> Could you try with:
> 
> ```shell
> npm i https://pkg.pr.new/atinux/nuxt-auth-utils@379
> ```
> 
> And confirm it fixes your issue?

Running some testing.

Tmmcmasters:
@atinux  Removed ssl required and ran into a security policy issue so I might not be able to test that. However, whenever I published my fork to a our own npm registry and tested with that, I was able to get it to work and it resolved the Typescript issues that we were facing.

Tmmcmasters:
@atinux Let me know if the above testing was sufficient

atinux (approved):

PR #382, opened 2025-03-24
feat: add Salesforce, Slack and Heroku OAuth providers

phof:
- [x] Adds Salesforce, Slack, Heroku providers
- [x] Supports `state` param verification using existing `handleState`, `handleInvalidState` utils
- [x] Tested in the playground


atinux:
Well, nice work and thank you!

PR #386, opened 2025-04-03
feat: add support for authorization parameters and ensure state param is included from query when present in Facebook & Microsoft OAuth

mendrinos:
- Added support for the `state` parameter in the Facebook OAuth configuration.
- Encoded the `state` parameter as a JSON string to allow custom data to be passed (e.g., redirect subdomain).


mendrinos:
Hi @atinux, Do you have a recommendation for the best way to define the "state" oauth parameter dynamically?

I’d like it to store dynamic JSON attributes such as csrf, redirect_deep_link, and host (especially in a multi-tenant subdomain setup). 

Ideally, I would want to construct this state object within a middleware, where I’m identifying the hostname and loading configuration in the TenantStore.

atinux (approved):

Issue #413, opened 2025-05-31
Keycloak: calls to endpoint userinfo should use `serverUrlInternal`


pkoz:
The OAuth module for Keycloak has a configuration parameter `serverUrlInternal` that is useful when the internal Keycloak URL (e.g., from the Nuxt server) is different from the public URL. 

The problem: When calling `/protocol/openid-connect/userinfo`, the module uses `serverUrl` instead of `serverUrlInternal` even though it is also an internal request.

The solution: a one-liner that replaces `serverUrl` with `serverUrlInternal`.

PR #414, opened 2025-05-31
bugfix: in Keycloak module, use `realmURLInternal` for internal calls

pkoz:
Solves issue #413


alexandre-hallaine:
Nice work

InjustFr:
Any news if this is going to be merged or not and if yes when ? I'm running into the same issue and I've had to hack this fix in so that I can have nuxt auth utils work with Keycloak

alexandre-hallaine:
This PR has been open for quite a while. At this point, the best option is probably to publish (npm) your own build with this fix applied.

atinux:
Actually it was already fixed in latest nuxt-auth-utils release https://github.com/atinux/nuxt-auth-utils/releases/tag/v0.5.22

PR #423, opened 2025-06-14
Add GitHub `state` param

tomlewis0:
Add support for the `state` param to be passed to the GitHub OAuth. This is required to prevent CSRF attacks.

`state` is accepted by GitHub and defined here: https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#1-request-a-users-github-identity

Similar to https://github.com/atinux/nuxt-auth-utils/pull/386 & https://github.com/atinux/nuxt-auth-utils/pull/192/files

### DIFF
--- a/src/runtime/server/lib/oauth/github.ts
+++ b/src/runtime/server/lib/oauth/github.ts
@@ -114,7 +114,7 @@ export function defineOAuthGitHubEventHandler({ config, onSuccess, onError }: OA
       authorizationParams: {},
     }) as OAuthGitHubConfig
 
-    const query = getQuery<{ code?: string, error?: string }>(event)
+    const query = getQuery<{ code?: string, error?: string, state?: string }>(event)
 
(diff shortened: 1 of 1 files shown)


src/runtime/server/lib/oauth/github.ts
```diff
@@ -144,11 +145,16 @@ export function defineOAuthGitHubEventHandler({ config, onSuccess, onError }: OA
...
           redirect_uri: redirectURL,
           scope: config.scope.join(' '),
+          state,
           ...config.authorizationParams,
         }),
       )
     }
 
+    if (query.state !== state) {
+      handleInvalidState(event, 'github', onError)
```
atinux, line 155:
```suggestion
      return handleInvalidState(event, 'github', onError)
```

atinux:
Thank you ❤️

PR #432, opened 2025-07-13
fix(LiveChat): Remove scope assumption with LiveChat login

mariosant:
This PR removes scope assumption with LiveChat login as it caused an issue when default scope would be necessary.
Also, fixes playground issues around LiveChat login.

### DIFF
--- a/playground/app/pages/index.vue
+++ b/playground/app/pages/index.vue
@@ -1,288 +1,288 @@
 <script setup lang="ts">
-const { user, openInPopup } = useUserSession();
+const { user, openInPopup } = useUserSession()
 
-const inPopup = ref(false);
+const inPopup = ref(false)
 const providers = computed(() =>
   [
     {
-      title: user.value?.google || "Google",
-      to: "/auth/google",
+      title: user.value?.google || 'Google',
+      to: '/auth/google',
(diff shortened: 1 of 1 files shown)


mariosant:
Breaking change is indeed an issue here but I don't think it makes much difference as scope is also defined within LiveChat console (where the app is being registered). If it's left outside it uses the default scope.

atinux:
I don't understand why you changed it to a string actually, why?

mariosant:
Hey @atinux, it's been a while and I can't recall very well tbh. I had some issues trying to apply default scope and I switched to string to make it as explicit and simple as possible. LiveChat applies default scope set by the app registration screen anyway, so most people won't even set it.

atinux:
All the scopes are defined as Array of string for consistency, so I suggest we keep them like before.

What do you recommend though is to make the default scope empty, and if empty, don't send the `scope` argument to let the default defined in LiveChat OAuth settings?

atinux:
Could you please try locally with my latest changes and confirm it's working @mariosant ?

mariosant:
Apologies for the delays @atinux, I will be confirming these soon!

mariosant:
Things work and this branch was updated with latest from `main` one!
Let's merge this @atinux !

Issue #309, opened 2024-12-15
Add Option to Disable Session Fetching During SSR


Latex88:
Is it possible to disable session fetching during SSR and fetch only on the client side? Currently, this behavior is achievable for routes that are cached or pre-rendered. However, it would be ideal to introduce a configurable setting that allows this behavior for non-cached routes as well, enabling client-only session fetching across all routes.

fredericDelaporte:
I really need this, because pre-rendering or swr does not fit all needs.

I want to be able to just directly emit the cache-control header in my publicly cacheable pages, but auth-utils put the current session in the page payload and causes session leakage. (I have made sure nothing call useUserSession during SSR on theses pages, but still, the session gets loaded.)

At that point I am stuck for fixing perf issues on my Web site due to not finding a way to disable session fetches during SSR on the pages I want to cache in my CDN. (And also because SWR and pre-rendering does not work with i18n, ruling theses possibilities out... https://github.com/nuxt-modules/i18n/issues/2342. Anyway, I also need to control the caching based on some runtime checks, not doable with route rules.)

I can put a workaround to force reloading the session, hiding the trouble, but that still lets information from the user having generated the cached page leak for those who analyze the SSR payload.

fredericDelaporte:
As a workaround, I wrote a `useHttpCache` composable taking care of resetting the auth session state. So, when needing to not have the session state in the SSR payload and have it loaded from the client side, I use this code:

```
if (import.meta.server) {
  // Hack: clear session state.
  useState('nuxt-session').value = null
  useState('nuxt-auth-ready').value = false
  // nuxt-auth-utils use this flag for loading the session only on mounted on client side,
  // for avoiding hydration mismatches.
  useNuxtApp().payload.isCached = true
}
```

It implies the server will still have needlessly loaded the session. And if the state is not cleared soon enough, and the page uses it, it will use the not cleared state.

The way the session is loaded by default, I do not see how I could prevent this dynamically on runtime from a page setup, where I can get the data I need to know whether the page should be publicly cached or not.

Full code of my `useHttpCache` composable:

```
/**
 * Emit a Cache-Control response header if the page is rendered on the server.
 * If public, will clear the user session state.
 * @param maxAge Maximum duration of cache freshness.
 * @param isPublic Should the response be publicly cacheable? false by default.
 */
export default function (maxAge: number, isPublic?: boolean) {
  maxAge = Math.floor(maxAge)
[...]

PR #435, opened 2025-07-14
feat: allow disabling session auto-loading

fredericDelaporte:
See #309


fredericDelaporte:
I have checked this with the playground.

I have attempted adding an additional test, but that was causing the tests to fail.

> Error: No context is available. (Forgot calling setup or createContext?)

Altering the exiting test to test my change did work, but better test the default use case, I think.

atinux:
Closing in favour of #443

PR #438, opened 2025-07-29
feat(zitadel) use more secure PKCE flow

julien1619:
This PR adds the PKCE flow to the Zitadel provider as it's recommended by the provider, inspired by what was done in `azureb2c` provider.


atinux:
By looking at https://oauth.net/2/pkce/ it seems that client secret is still recommend and PKCE is not replacing client secret, could you please add it back or justify why we can remove it @julien1619 ?

julien1619:
Ok I will add it back but as an optionnal value because in Zitadel if you choose PKCE you won't have any secret generated.

Here is the related issue in Zitadel: https://github.com/zitadel/zitadel/issues/7110

atinux:
Please mention me when ready @julien1619

julien1619:
@atinux It seems Ok for me now.

atinux (approved):
Thank you!

PR #444, opened 2025-08-19
feat: add OpenID Connect (OIDC) provider

### LINKED ISSUE #89: Support for OIDC providers which expose `.well-known/openid-configuration`
I was sad to see that the closest thing currently available is the Keycloak provider however that mandates usage of a realm. Many OIDC providers providers, especially FOSS ones which you can self-host provide a standardized `well-known/openid-configuration` endpoint where all further endpoints and supported values are exposed. It would be great to support this and would eliminate the need for many specialized providers. Ideally, there is a general manualOIDC provider where one can manually set the authorization/token/userinfo/revokation etc endpoint and one wellKnownOIDC provider which simply takes a single URL, fetches the values and delegates the rest to the manualOIDC provider.

larsrickert:
closes #25
closes #89

This PR adds support for a generic OpenID connect (OIDC) provider that can be used with any provider that supports the OIDC standard.

It supports the 'code' response type, grant type 'authorization_code' and PKCE (so now client secret is needed).

Since the existing PR #25 has not seen progress in almost two years, this PR is intended to replace #25.

## Open questions

@maintainers Please leave your feedback for the following questions:

- Should also a non-PKCE flow be supported that uses a client secret instead?

### DIFF
--- a/src/runtime/server/lib/oauth/oidc.ts
+++ b/src/runtime/server/lib/oauth/oidc.ts
@@ -14,12 +14,6 @@ export interface OAuthOidcConfig {
    */
   clientId?: string
   /**
-   * OAuth Client Secret
-   *
-   * @default process.env.NUXT_OAUTH_OIDC_CLIENT_SECRET
-   */
-  clientSecret?: string
-  /**
    * URL to the OpenID Configuration endpoint. Used to fetch the endpoint URLs from.
    *
    * @default process.env.NUXT_OAUTH_OIDC_CONFIG_URL
@@ -247,11 +241,11 @@ export function defineOAuthOidcEventHandler<TUser = OidcUser>({ config, onSucces
       return onError(event, error)
     }
 
-    if (!config.clientId || !config.clientSecret || !config.configUrl) {
-      return handleMissingConfiguration(event, 'oidc', ['clientId', 'clientSecret', 'configUrl'], onError)
+    if (!config.clientId || !config.configUrl) {
+      return handleMissingConfiguration(event, 'oidc', ['clientId', 'configUrl'], onError)
     }
 
-    const oidcConfig = await $fetch<{ authorization_endpoint: string, token_endpoint: string, userinfo_endpoint: string }>(config.configUrl)
+    const oidcConfig = await $fetch<{ authorization_endpoint: string, token_endpoint: string, userinfo_endpoint?: string }>(config.configUrl)
 
     const redirectURL = config.redirectURL || getOAuthRedirectURL(event)
     const state = await handleState(event)
@@ -282,7 +276,6 @@ export function defineOAuthOidcEventHandler<TUser = OidcUser>({ config, onSucces
       body: {
         grant_type: 'authorization_code',
         client_id: config.clientId,
-        client_secret: config.clientSecret,
         redirect_uri: redirectURL,
         code: query.code,
         code_verifier: verifier?.code_verifier,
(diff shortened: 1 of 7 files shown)


src/runtime/server/lib/oauth/oidc.ts
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    const tokens = await requestAccessToken<OidcTokens & { error?: unknown }>(oidcConfig.token_endpoint, {
+      body: {
+        grant_type: 'authorization_code',
+        client_id: config.clientId,
+        client_secret: config.clientSecret,
+        redirect_uri: redirectURL,
+        code: query.code,
+        code_verifier: verifier?.code_verifier,
```
randriese, line 300: this part does not seem to omit itself when a clientSecret is provided, resulting in a 400 Bad Request error when fetching the tokens.

larsrickert: In line 269 the verifier is set to undefined if a client secret is passed so its omitted in this case.

For me using a client secret works as expected 🤔  Which OIDC provider are you using?

larsrickert: Thanks for testing! I updated the implementation to only include the PKCE query parameters if the client ID is unset.
Would you mind testing it again?

Natronick:
Hi @larsrickert ! Great work.
Are you considering the possibility of retrieving user data by validating and decoding the ID token if it's provided by the provider (alongside the Access token)?

randriese:
> Hi @larsrickert ! Great work.
> 
> Are you considering the possibility of retrieving user data by validating and decoding the ID token if it's provided by the provider (alongside the Access token)?

That's already implemented :)

src/runtime/server/lib/oauth/oidc.ts
randriese: Works as expected now!

Natronick:
> > Hi @larsrickert ! Great work.
> > Are you considering the possibility of retrieving user data by validating and decoding the ID token if it's provided by the provider (alongside the Access token)?
> 
> That's already implemented :)

Thanks for the quick reply! Could you point me to where this is implemented? I'm not seeing it in the PR code. Currently, it seems we only retrieve user info via the userinfo endpoint (which is optional), and the OidcTokens interface doesn't contain the idToken. Maybe I'm missing something?

randriese:
> > > Hi @larsrickert ! Great work.
> 
> > > Are you considering the possibility of retrieving user data by validating and decoding the ID token if it's provided by the provider (alongside the Access token)?
> 
> > 
> 
> > That's already implemented :)
> 
> 
> 
> Thanks for the quick reply! Could you point me to where this is implemented? I'm not seeing it in the PR code. Currently, it seems we only retrieve user info via the userinfo endpoint (which is optional), and the OidcTokens interface doesn't contain the idToken. Maybe I'm missing something?

Maybe I misunderstood? As far a I saw during testing the user data is decoded in this server route which is used when redirecting from the oidc provider back to the Frontend.

playground/server/routes/auth/oidc.ts

larsrickert:
> Maybe I misunderstood? As far a I saw during testing the user data is decoded in this server route which is used when redirecting from the oidc provider back to the Frontend.

Yes the user data is retrieved by sending the access token to the userinfo endpoint of the OIDC provider (thats already implemented). There is also a second way how the user data could be retrieved (e.g. because some providers do not offer a userinfo endpoint) and thats by requesting an idtoken (JWT with user data encoded in it) that is then validated and decoded.

I'd suggest to aleady merge the OIDC provider as it is so its not blocked by another optional feature :) We can then extend it if needed in follow up PRs

larsrickert:
@randriese Thanks for your time reviewing the PR! Is there anything left in order to merge this PR?
I wish you already a great christmas time and a happy new year!

randriese (approved):

randriese:
@larsrickert looks good to me! thanks for your efforts in bringing OIDC to the package!
happy holidays and all the best for 2026!

WiRight:
any update? this PR can help me with casdoor intergrate

atinux:
Thanks for this great PR @larsrickert! 🙏

I've pushed a small fix commit (b678617) addressing a few issues I spotted during review:

### Changes made:

1. **Fixed userinfo endpoint request** - The userinfo endpoint was incorrectly using `body` parameter, but OIDC userinfo is typically a GET request. Changed to spread custom parameters into headers instead.

2. **Fixed config casing** - Changed `redirectUrl` to `redirectURL` in `module.ts` to match the interface definition and be consistent with other providers.

3. **Enhanced OidcTokens interface** - Added missing standard OIDC token response fields:
   - `id_token` - Allows consumers to decode the ID token themselves if needed (addresses the discussion in comments)
   - `refresh_token` - Standard field for token refresh flows
   - `expires_in` - Token expiration time
   - Made `scope` optional since not all providers return it

The PR looks great overall - the PKCE implementation is solid and follows existing patterns well. Ready to merge! 🚀

larsrickert:
@atinux Looks great, thanks! As @marijngiesen commented in https://github.com/atinux/nuxt-auth-utils/pull/444#discussion_r2686181540, there is a small typo in the "OAuthProvider" type. It should be "oidc" instead of "oicd" :) 

Edit: I created a small fix PR: #485

PR #463, opened 2025-10-11
fix: add explicit security options to OAuth cookies to prevent state mismatch errors

bisand:
### Fix: Add proper security options to OAuth state and PKCE cookies
The Problem
The OAuth state and PKCE verifier cookies were being set without any explicit options, which caused two main issues:

#### Security vulnerabilities:
- Cookies were accessible to JavaScript (XSS risk)
- Cookies could be sent over HTTP (man-in-the-middle risk)
- No CSRF protection

#### Browser compatibility:
- Modern browsers have started enforcing stricter cookie policies by default
- Without explicit sameSite settings, some browsers drop cookies during OAuth redirects
- _**This caused "state mismatch" errors, especially with providers like Okta**_

#### What This PR Does:
Adds explicit cookie security options to both handleState() and handlePkceVerifier():

```typescript
setCookie(event, 'nuxt-auth-state', state, {
  httpOnly: true,
  secure: process.env.NODE_ENV !== 'development',
  sameSite: 'lax',
  maxAge: 60 * 10, // 10 minutes
  path: '/',
})
```

#### Why these specific options:

- `httpOnly: true` - Prevents JavaScript access, blocks XSS attacks
- `secure: true` in production - HTTPS only when it matters
- `sameSite: 'lax'` - Allows cookies during OAuth redirects while preventing CSRF
- `maxAge: 600` - OAuth flows should complete within 10 minutes
- `path: '/'` - Available across the entire application

#### Impact
This affects all OAuth providers in the library (GitHub, Google, Microsoft, Azure B2C, Okta, Auth0, Keycloak, etc.) since they all use these utility functions.

No breaking changes - the functions work exactly the same, just more securely.

Why `!== 'development'` instead of `=== 'production'`
Using the negative check means if NODE_ENV is undefined or misspelled, cookies default to secure mode. Better to have secure cookies break local dev (easy to fix) than accidentally ship insecure cookies to production.

**Testing**
Tested OAuth flows with several providers in both dev and production modes. The state mismatch errors are gone, and cookies are properly secured.

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,7 +597,7 @@ This is because the user session is stored in a secure cookie and cannot be acce
 
 **This means that you should not rely on the user session during prerendering.**
 
-You may also choose to instruct Nuxt AUth Utils to fetch the user session only on the client side, with the `loadStrategy` option in your `nuxt.config.ts`:
+You may also choose to instruct Nuxt Auth Utils to fetch the user session only on the client side, with the `loadStrategy` option in your `nuxt.config.ts`:
 
 ```ts
 export default defineNuxtConfig({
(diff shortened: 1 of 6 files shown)


atinux (approved):

atinux:
Thank you so much @bisand

PR #465, opened 2025-11-06
fix: missing code_challenge and state mismatch

jals1212:
### FIX: Update handlePkceVerifier to prevent missing code_challenge, and prevent state mismatch.

**Description:**

If something goes wrong in authorize endpoint (some extra authorizationParams missing or incorrect) and you try again to authorize you end up with a missing code_challenge in the query, same for state (state mismatch).

**Fix:**
Read the code (from event query) inside the handlePkceVerifier and handleState functions, if the code is present then  is not  necessary to regenerate the code_challenge or the state.

### DIFF
--- a/src/runtime/server/lib/oauth/azureb2c.ts
+++ b/src/runtime/server/lib/oauth/azureb2c.ts
@@ -105,7 +105,6 @@ export function defineOAuthAzureB2CEventHandler({ config, onSuccess, onError }: 
       return handleInvalidState(event, 'azureb2c', onError)
     }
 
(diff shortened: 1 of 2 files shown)


atinux:
Why not reading the `query.code` directly from the `event` in the `handlePkceVerifier` and `handleState` utils? @jals1212

jals1212:
> Why not reading the `query.code` directly from the `event` in the `handlePkceVerifier` and `handleState` utils? @jals1212 

You're right that's less invasive and leaner, I'll make the change. Thanks.

jals1212:
Ready @atinux

Issue #450, opened 2025-09-05
Opt-out of session plugin


IlyaSemenov:
I would like to be able to opt-out of forced client-side session fetch that happens on each page load:

https://github.com/atinux/nuxt-auth-utils/blob/0ff5cc7aeef3081ea24de68d3ef9d3030ff3c9a7/src/module.ts#L100
https://github.com/atinux/nuxt-auth-utils/blob/0ff5cc7aeef3081ea24de68d3ef9d3030ff3c9a7/src/runtime/app/plugins/session.client.ts#L6-L8

I have two specific use cases in mind for avoiding the default plugin:

1. I would like to optimize no-SSR data fetch as explained here #449 (if you decide that approach is too magicky to land in the core, at least allow this to be implemented with a user-land plugin).
2. For some projects, I also would like to deliver non-JSON-serializable user data. For instance, I have a project involving e2e encryption, where the auth session brings WebCrypto keys (Uint8Array's). Most of the client-server communication is performed via [tRPC](https://trpc.io/) and a custom serializer that passes binary data transparently, but specifically for the auth session I had to come up with ugly boilerplate and intermediary types. I would like to either fetch the session data with a custom tRPC endpoint, or perhaps combine this with (1) and deliver this with a custom Nuxt payload.

How I see it, it will be a simple option in `nuxt.config.ts`:

```ts
export default defineNuxtConfig({
  session: {
    plugin: false // or injectPlugin: false
  }
})
```

I can come up with a PR if you generally approve this.

atinux:
Like commented in #472, let's have `loadStrategy: 'none'` to disable.

IlyaSemenov:
I see it has been addressed in #472, thanks.

PR #472, opened 2025-11-20
feat: opt out of session plugin

samydoesit:
Hello,

this PR adds the ability to disable the session plugin.

In our specific use case we don't rely on the session utilities provided.
Unlike the scenario described in #450, we simply want to deactivate the session plugin to avoid unnecessary overhead.

This change introduces an option to disable the plugin and these unwanted session fetches.

Let me know what you think!


atinux:
I would prefer to improve the `loadStrategy` to have the `none` option to disable the plugins instead @samydoesit

PR #475, opened 2025-12-01
fix(zitadel): don't prefix domain with `https://` if another protocol is used

ysemennikov:
Currently, it's impossible to use non-https base URL in the Zitadel connector, since the connector always prefixes the domain with `https://`.

This is very inconvenient when testing the setup locally, since locally we usually don't use https.

The PR fixes the issue.


ysemennikov:
Hi @atinux, there was a TS issue indicated by the type-checking pipeline. Just fixed it.

lennartzellmer:
@ysemennikov Thanks so much for fixing this! (I just ran into this issue and saw that you fixed it, and I'm so happy I don't have to do it now :D )

PR #497, opened 2026-02-02
feat: OIDC support for jwks

hareland:
This PR aims to add support for using JWKS to verify the token if provided in config or OIDC endpoint..

### DIFF
--- a/src/runtime/server/lib/oauth/oidc.ts
+++ b/src/runtime/server/lib/oauth/oidc.ts
@@ -7,6 +7,7 @@ import type { QueryObject } from 'ufo'
 import { withQuery } from 'ufo'
 import type { RequestAccessTokenBody } from '../utils'
 import { verifyJwt, getOAuthRedirectURL, handleAccessTokenErrorResponse, handleInvalidState, handleMissingConfiguration, handlePkceVerifier, handleState, requestAccessToken, handleNonce, parseJwt } from '../utils'
+import type { JWTPayload } from 'jose'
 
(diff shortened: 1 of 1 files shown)


src/runtime/server/lib/oauth/oidc.ts
hareland: Since not all OIDC providers support the `/userinfo` endpoint, this should probably be inferred from the `ID Token` **IF** it is verified?

larsrickert:
Hey, one question regarding the aim of this PR:
Why do you want to very the same token that you have just received from the IDP (when logging in) with the exact same IDP?

Example: You login via GitHub and get the tokens in the response from GitHub. With the changes in this PR, we would immediately verify / ask GitHub again "Are these tokens valid?"
Since we just received them, do you have any specific use case in mind why this is needed?

hareland:
Hi @larsrickert, the purpose is to independently verify the signature of the JWT using the provider's public key.

- [This is a feature of OIDC](https://mojoauth.com/glossary/oidc-json-web-key-endpoint/)
- [Not all OIDC providers implement the `/userinfo` endpoint.](https://mojoauth.com/glossary/oidc-userinfo-endpoint/)

As it stands now, this will only verify the signature IF there is JWKS defined in the `.openid-configuration`, this could be a different server.

FYI: This PR is not complete, and i am looking for input so any suggestions are more than welcome ❤️ 

---
Edit: This should probably be an **opt-in** feature

PR #501, opened 2026-02-12
feat: use nuxt-site-config module if available for origin extraction

### LINKED ISSUE #500: [FEATURE]: Better integration for redirect urls
Is it possible for this module to use the `site-config` module to guess the url first if the user has it installed?

https://nuxtseo.com/docs/site-config/getting-started/installation

astraldev:
Fixes #500 

I've added an extra plugin to check if nuxt site config is available, and use it's `getNitroOrigin` composable at the `getOAuthRedirectURL` method

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -102,6 +102,7 @@ export default defineNuxtModule<ModuleOptions>({
     }
     // Server
     addServerPlugin(resolver.resolve('./runtime/server/plugins/oauth'))
+    addServerPlugin(resolver.resolve('./runtime/server/plugins/site-config'))
     addServerImportsDir(resolver.resolve('./runtime/server/lib/oauth'))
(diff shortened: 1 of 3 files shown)


atinux:
Not sure I am happy with this approach, most people use the `NUXT_PUBLIC_SITE_URL` env variable, worst case we can read from it

astraldev:
Understandable, if the module itself can have support for a base url, then there would be no need for it. I picked it since the `getNitroOrigin` is a well feature packed utility. 

Would you prefer I clone it instead (only the relevant part)? So we don't rely on that?

It's quite strange, we have to configure the redirect url for each oauth provider manually.

Using the  `NUXT_SITE_PUBLIC_URL` would still mean having to manually configure each of them again since the guess method doesn't really do so well

atinux:
But my main concern is why the current behavior does not work for you @astraldev?

PR #502, opened 2026-02-15
docs: fix passwordNeedsRehash usage in README

richex-cn:
The `passwordNeedsRehash` function is synchronous, but the example showed it being used with 'await'. This commit removes the misleading keyword to align with the actual implementation.

https://github.com/atinux/nuxt-auth-utils/blob/v0.5.28/src/runtime/server/utils/password.ts#L46-L64


atinux:
thank you!

PR #504, opened 2026-02-17
fix: Keycloak provider unhandled error on requestAccessToken

### LINKED ISSUE #503: Keycloak provider unhandled error on requestAccessToken
I think this fetch should be handled, because apparently an unhandled error can happen here:
https://github.com/atinux/nuxt-auth-utils/blob/main/src/runtime/server/lib/oauth/keycloak.ts#L116

Image

Probably something along the lines of:
```ts 
const tokens = await requestAccessToken(tokenURL, {
      body: {
        grant_type: 'authorization_code',
        client_id: config.clientId,
        client_secret: config.clientSecret,
        redirect_uri: redirectURL,
        code: query.code,
      } })
// Handle the error
.catch((error) => {
    if (error instanceof FetchError) {
      return handleAccessTokenErrorResponse(event, 'keycloak', error.data, onError)
    }
    return handleAccessTokenErrorResponse(event, 'keycloak', error, onError) 
})
```

letoast:
resolves #503

### DIFF
--- a/src/runtime/server/lib/oauth/keycloak.ts
+++ b/src/runtime/server/lib/oauth/keycloak.ts
@@ -123,10 +123,10 @@ export function defineOAuthKeycloakEventHandler({
         code: query.code,
       } })
       .catch((error) => {
-          if (error instanceof FetchError) {
-            return handleAccessTokenErrorResponse(event, 'keycloak', error.data, onError)
-          }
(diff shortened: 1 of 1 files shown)


letoast:
thanks for the description edit @atinux , I'm not much of a contributor, sorry :D

PR #517, opened 2026-04-18
fix(oidc): dedupe 'openid' scope in config

### LINKED ISSUE #516: OIDC provider: duplicate 'openid' scope due to defu array merging
When using defineOAuthOidcEventHandler with a config that already includes 'openid' in the scope array, the resulting scope contains 'openid' twice due to how `defu` merges arrays.

https://github.com/atinux/nuxt-auth-utils/blob/ceb366b629130a2da539750a86b6cccdcc70ae00/src/runtime/server/lib/oauth/oidc.ts#L246-L248
`defu` concatenates arrays rather than replacing them. So if the input config is `{ scope: ['openid', 'profile'] }`, the merged result becomes `{ scope: ['openid', 'profile', 'openid'] }`.

This produces a scope string of "openid profile openid" in the authorization request, which may cause errors with some OIDC providers that reject duplicate scopes.

This is the same class of issue reported in unjs/defu#136, which was closed as "not planned" since deduplication is left to consumers.

## Reproduction

```
defineOAuthOidcEventHandler({
  config: {
    scope: ['openid', 'profile', 'email'],
  },
  async onSuccess(event, { user, tokens }) {
    // ...
  },
})
```

The authorization redirect will contain scope=openid+profile+email+openid.

tmlmt:
Fixes: #516

## Changes

Replaced the `defu` default `[{ scope: ['openid'] }]` with an explicit check that ensures 'openid' is present without duplication, consistent with how the Dropbox provider already handles it:

https://github.com/atinux/nuxt-auth-utils/blob/ceb366b629130a2da539750a86b6cccdcc70ae00/src/runtime/server/lib/oauth/dropbox.ts#L78-L80

### DIFF
--- a/src/runtime/server/lib/oauth/oidc.ts
+++ b/src/runtime/server/lib/oauth/oidc.ts
@@ -243,9 +243,10 @@ interface OIDCConfiguration {
  */
 export function defineOAuthOidcEventHandler<TUser = OidcUser>({ config, onSuccess, onError }: OAuthConfig<OAuthOidcConfig, { user: TUser, tokens: OidcTokens }>) {
   return eventHandler(async (event: H3Event) => {
-    config = defu(config, useRuntimeConfig(event).oauth.oidc, {
-      scope: ['openid'],
-    } satisfies OAuthOidcConfig)
+    config = defu(config, useRuntimeConfig(event).oauth.oidc)
+    if (!config.scope?.includes('openid')) {
+      config.scope = [...(config.scope || []), 'openid']
+    }
 
     const query = getQuery<{ code?: string, error?: string, state?: string }>(event)
 


ImBoop:
It should probably actually just create a fresh object or only merge the config once. the `config = defu(config,...)` is modifying the original object every request so one could instead do `config = defu({}, ...)` instead (or again just do the merging beforehand such as before the `return`)

tmlmt:
Interesting point but which should be raised in a separate issue as it is unrelated to the issue this PR is addressing.

ImBoop:
> Interesting point but which should be raised in a separate issue as it is unrelated to the issue this PR is addressing.

It is related. The cause is overwriting config every request w/ defu (the first parameter being the object target) - if the user specifies 'openid' in their scopes list, your deduplication only prevents it from being added an additional time. The cleaner fix is to build it before the handler at all, which would fix your issue as well as the general duplicated scopes list issue.

Your bug is caused by the defu duplication; just applying your fix without the additional nuances I mentioned will only fix one particular instance of the broader problem; the problem will continue to occur.

tmlmt:
What this PR fixes is a specific bug where 'openid' ends up duplicated in the scope list. The fix correctly replaces the defu default { scope: ['openid'] } approach with an explicit guard that only adds 'openid' if it's not already present. This directly and correctly solves issue #516.

What you are raising is a broader concern related to the fact that config = defu(config, ...) mutates the original config object on every request (since defu's first argument is the mutation target) but this does not address the target issue for this PR and would be worth elaborating in a separate issue.

PR #521, opened 2026-05-03
fix(sessison): safe usage of localStorage (#519)

### LINKED ISSUE #519: Safe usage of localStorage?
Hi,

I've noticed that the plugin **src/runtime/app/plugins/session.client.ts** breaks during mounting, when localStorage is not available, for instance, when the user block cookies (which seems to forbid the use of localStorage too).

In that case, I have a blank page, with the following error:
> nuxt] error caught during app initialization SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.

When I remove [this block](https://github.com/atinux/nuxt-auth-utils/blob/v0.5.29/src/runtime/app/plugins/session.client.ts#L18-L23), the error does not occur anymore:

```ts
if (localStorage.getItem('temp-nuxt-auth-utils-popup')) {
  // There is a local storage item. That's mean we are coming back in the popup
  localStorage.removeItem('temp-nuxt-auth-utils-popup')
  const error = useError()
  if (!error.value) window.close()
}
```

Could it be possible to check localStorage availability before using it? With either a simple try/catch or the [MDN suggested approach](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability)?

I know cookies are mandatory for the auth flow to work properly, but I think we should have a chance to display a meaningful error message or error page, instead of breaking the app.

[Here](https://github.com/search?q=repo%3Aatinux%2Fnuxt-auth-utils%20localStorage&type=code) are the calls to localStorage.

Thanks!

HugoMuller:
Wrap calls to `localStorage` in try/catch, so that the app does not crash during the app mounting.

Fixes #519

### DIFF
--- a/src/runtime/app/composables/session.ts
+++ b/src/runtime/app/composables/session.ts
@@ -45,8 +45,13 @@ export function useUserSession(): UserSessionComposable {
     }
   }
   const openInPopup = (route: string, size: { width?: number, height?: number } = {}) => {
-    // Set a local storage item to tell the popup that we pending auth
-    localStorage.setItem('temp-nuxt-auth-utils-popup', 'true')
+    try {
+      // Set a local storage item to tell the popup that we pending auth
+      localStorage.setItem('temp-nuxt-auth-utils-popup', 'true')
+    }
+    catch (error) {
+      console.error('[nuxt-auth-utils] Could not use localStorage during popup flow initialization.', error)
+    }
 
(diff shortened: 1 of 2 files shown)


HugoMuller:
Hi @atinux, any chance to get a review on this PR? 🙏

Issue #488, opened 2026-01-14
Swallowed fetch failure


fredericDelaporte:
The user session `fetch` method appears to have a "catch all" that swallows all troubles silently, then pretends the session is ready.

https://github.com/atinux/nuxt-auth-utils/blob/f55debb6b8c16e38401f2f827068a5b4d8eca2ed/src/runtime/app/composables/session.ts#L29-L39

That is troublesome in case of failure, since the application user code has no way of detecting it, and no way of reporting the trouble.

I have searched the file history and saw it was there since the creation of the file, so, I found no explanation for this in the source. I have also searched issues without finding any about this.

What is the rationale for this implementation?

Should it be changed?

Underlying trouble that caused me to see this: I have implemented a "refresh" of the user session data by calling this `fetch` periodically (through `setTimeout`). It is meant for detecting if it has been changed through actions on another tab of the browser. When a fetch failure occurs, it causes the user data to switch to `null`, causing the Web site to switch to "unauthenticated" mode, while actually, the authentication is still there: reloading entirely the page allows it to switch back to logged in.

More troublesome to me, I have no way to see the fetch has actually failed to diagnose and fix the trouble, or at least, handle some retry logic. (In the browser, in my case, the trouble seems the HTTP request being sometime cancelled, appearing in the browser network pane without any response. No idea what would cause that cancellation.)

fredericDelaporte:
I would rather have the fetch letting the failure bubble up, but that could be a breaking change.

Otherwise, would it be possible to preserve the current data, and expose some way to retrieve the failure if any?

Something like:

```js
  const fetch = async () => {
    try {
      sessionState.value = await useRequestFetch()<UserSession>('/api/_auth/session', {
        headers: {
          accept: 'application/json',
        },
        retry: false,
      })
    } catch (err) {
      fetchError.value = err
      return
    }
    fetchError.value = undefined

    if (!authReadyState.value) {
      authReadyState.value = true
    }
  }
```

`fetchError` would be added on UserSession.

atinux:
The silent catch is not ideal here, the catch was mostly during SSR to avoid breaking the app is something unexpected is sent back from the server (for instance, missing `NUXT_SESSION_PASSWORD` env variable).



I think we could expose a `fetchState`, but first, for what reason your `fetch` is failing?

fredericDelaporte:
[...]

PR #532, opened 2026-06-18
fix: log fetch failure

fredericDelaporte:
Help diagnose #488

## Summary by CodeRabbit

* **Improvements**
  * Enhanced error logging for session request failures to aid troubleshooting.
  * Session handling remains unchanged when requests fail, while failure details are now recorded for easier diagnosis.

### DIFF
--- a/src/runtime/app/composables/session.ts
+++ b/src/runtime/app/composables/session.ts
@@ -32,7 +32,7 @@ export function useUserSession(): UserSessionComposable {
         accept: 'application/json',
       },
       retry: false,
-    }).catch(reason => {
+    }).catch((reason) => {
       console.log('Failure on fetching session', reason)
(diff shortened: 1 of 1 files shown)


src/runtime/app/composables/session.ts
```diff
@@ -32,7 +32,10 @@ export function useUserSession(): UserSessionComposable {
         accept: 'application/json',
       },
       retry: false,
-    }).catch(() => null)
+    }).catch(reason => {
+      console.log('Failure on fetching session', reason)
+      return null
```
fredericDelaporte, line 37: We may somewhat fix the issue by yielding `sessionState.value` here instead of `null`. It would allow to preserve the session state on fetch failure, instead of losing it.

